PR #24, opened 2022-08-08
Change Example app configuration and README

descorp:
# Open PR

## Changes

* remove all pass-through params from `AdyenCheckoutContext.js`
* move all non-SDK configurations to `Example/APIClient.js`
* Update README
* Add documentation folder and `Configuration` doc

### DIFF
--- a/example/src/APIClient.js
+++ b/example/src/APIClient.js
@@ -1,20 +1,38 @@
-import { environment } from './Configuration';
+import { environment, channel } from './Configuration';
+
+const serverConfiguration = {
+  channel: channel,
+  shopperReference: 'Checkout Shopper',
+  reference: 'React Native'
+};
+
+const parseConfig = (configuration) => {
+    return {
+      merchantAccount: configuration.merchantAccount,
+      countryCode: configuration.countryCode,
+      shopperLocale: configuration.shopperLocale,
+      amount: configuration.amount
+  }
+};
 
 export const fetchPaymentMethods = (configuration) => {
   let body = {
-    merchantAccount: configuration.merchantAccount,
-    countryCode: configuration.countryCode,
-    shopperLocale: configuration.shopperLocale,
-    amount: configuration.amount,
-    channel: configuration.channel,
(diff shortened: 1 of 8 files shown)


example/src/APIClient.js
```diff
@@ -1,20 +1,38 @@
...
-  return fetchFrom(environment.url + 'payments', data);
+export const fetchPayments = (data, configuration) => {
+  let body = {
+    ...data,
+    ...parseConfig(configuration),
+    ...serverConfiguration,
+    additionalData: { allow3DS2: true }
+  };
+
+  return fetchFrom(environment.url + 'payments', body);
```
jesujcastillom, line 35: You may consider using [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)
```suggestion
  return fetchFrom(`${environment.url }payments`, body);
```

example/src/APIClient.js
```diff
@@ -1,20 +1,38 @@
...
+};
+
+const parseConfig = (configuration) => {
+    return {
+      merchantAccount: configuration.merchantAccount,
+      countryCode: configuration.countryCode,
+      shopperLocale: configuration.shopperLocale,
+      amount: configuration.amount
+  }
+};
```
jesujcastillom, line 16: You can use an implicit return in this case

```suggestion
const parseConfig = (configuration) => ({
    merchantAccount: configuration.merchantAccount,
    countryCode: configuration.countryCode,
    shopperLocale: configuration.shopperLocale,
    amount: configuration.amount
});
```

And given that the properties are called the same, you can use shorthand property names
```suggestion
const parseConfig = ({ merchantAccount, countryCode, shopperLocale, amount }) => ({
    merchantAccount,
    countryCode,
    shopperLocale,
    amount
});
```

ribeiroguilherme (dismissed):

ribeiroguilherme (approved):

jesujcastillom (approved):

PR #28, opened 2022-08-23
Update Android to 4.7.1

descorp:
# Open PR

## Changes

* update Anroid to 4.7.1
* convert AdyenCardComponent.java -> AdyenCardComponent.kt
* convert AdyenDropInComponent.java -> AdyenDropInComponent.kt
* convert BaseModule.java ->  BaseModule.kt
* add voucher action handling


android/src/main/java/com/adyenreactnativesdk/component/AdyenCardComponent.kt
```diff
@@ -0,0 +1,170 @@
...
+        val actionHandlerConfiguration = ActionHandlerConfiguration(shopperLocale, environment, clientKey)
+        actionHandler = ActionHandler(this, actionHandlerConfiguration)
+
+        val parser = CardConfigurationParser(configuration)
+        val componentConfiguration: CardConfiguration
+        val builder = CardConfiguration.Builder(shopperLocale, environment, clientKey)
+        componentConfiguration = parser.getConfiguration(builder)
+
+        val theActivity = appCompatActivity
+        val viewModel = ComponentViewModel(paymentMethod!!, shopperLocale, amount)
```
OscarSpruit, line 69: Be careful with `!!`, it will throw an exception when the object is `null`. It is better to handle these cases safely

descorp: Updated! Thanks :)

android/src/main/java/com/adyenreactnativesdk/ActionHandler.kt
```diff
@@ -53,13 +53,7 @@ class ActionHandler(
...
 ) : Observer<ActionComponentData> {
 
-    companion object {
-        val TAG = LogUtil.getTag()
-        const val UNKNOWN_ACTION = "UNKNOWN ACTION"
-        const val ACTION_FRAGMENT_TAG = "ACTION_DIALOG_FRAGMENT"
-        private const val BUNDLE_ACTION = "bundle_action"
-    }
-
+    private var dialog: DialogFragment? = null
```
OscarSpruit, line 56: It's better to not hold a reference to any kind of `Fragment`s and `Activity`s. If the referenced `Fragment` gets destroyed (because it's no longer in view), but this class still holds a reference to it, then the garbage collector is not able to collect it and thus there is a memory leak.

I believe you only have this reference to dismiss the dialog later. Dismissing a dialog can best be done by the parent `Fragment` or `Activity`

descorp: Fixed :)

OscarSpruit (approved):

PR #29, opened 2022-09-01
Instant pm

descorp:
# Open PR

## Changes

* make native components work via DropIn
* add instant payment
* improve example's PaymentMethodsView
* update iOS to 4.8.0


ios/Components/InstantComponent.swift
```diff
@@ -0,0 +1,155 @@
...
+    override static func requiresMainQueueSetup() -> Bool { true }
+    override func stopObserving() {}
+    override func startObserving() {}
+    override func supportedEvents() -> [String]! { super.supportedEvents() }
+
+    @objc
+    func hide(_ success: NSNumber, event: NSDictionary) {
+        DispatchQueue.main.async {
+            self.currentComponent?.finalizeIfNeeded(with: success.boolValue)
+            UIApplication.shared.keyWindow?.rootViewController?.dismiss(animated: true)
```
mohammedDehairy, line 30: just grabbing the `rootViewController`? :-D , that's a bit dangerous, because it will dismiss everything even if they don't belong to the SDK

ios/Components/InstantComponent.swift
```diff
@@ -0,0 +1,155 @@
...
+            self.actionHandler = nil
+            self.currentComponent = nil
+            self.currentPaymentComponent = nil
+        }
+    }
+
+    @objc
+    func open(_ paymentMethods: NSDictionary, configuration: NSDictionary) {
+        let paymentMethod: PaymentMethod
+        guard let data = try? JSONSerialization.data(withJSONObject: paymentMethods, options: []) else {
```
mohammedDehairy, line 41: why `JSONSerialization `?

ios/Components/InstantComponent.swift
```diff
@@ -0,0 +1,155 @@
...
+           let firstPaymentMethod = paymentMethods.regular.first {
+            paymentMethod = firstPaymentMethod
+        }
+        else if let jsonPaymentMethod = try? JSONDecoder().decode(AnyPaymentMethod.self, from: data),
+                let anyPaymentMethod = jsonPaymentMethod.value {
+            paymentMethod = anyPaymentMethod
+        }
+        else {
+            return assertionFailure("InstantComponent: Can not parse payment method.")
+        }
```
mohammedDehairy, line 55: also this code seems similar to `AdyenDropIn.open()`, can we put somewhere to be reusable?

ios/Components/InstantComponent.swift
```diff
@@ -0,0 +1,155 @@
...
+
+}
+
+extension InstantComponent: PresentationDelegate {
+
+    private static var presenter: UIViewController? { UIApplication.shared.keyWindow?.rootViewController }
+
+    func present(component: PresentableComponent) {
+        DispatchQueue.main.async {
+            self.present(component)
```
neelSharma12, line 97: maybe weakify self here?

example/src/PaymentMethodsView.js
```diff
@@ -18,43 +18,71 @@ const styles = StyleSheet.create({
...
     backgroundColor: isDarkMode ? Colors.black : Colors.white,
   };
   const platformSpecificPayment =
     Platform.OS === 'ios' ? 'Apple Pay' : 'Google Pay';
+  const platformSpecificType =
+    Platform.OS === 'ios' ? 'applepay' : 'googlepay';
+  const isAvailable = (type) => { 
+    if (!paymentMethods) { return false; }
+    return paymentMethods.paymentMethods.find(x => x.type === type.toLowerCase()) != null
+  };
```
ribeiroguilherme, line 34: Better practice is to wrap functions in `useCallback` hook, so they don't get recreated during the re-rendering step.

Also, you can use `Array.prototype.some` to verify if the payment method is in the array, so you don't have to do the check `!== null` at the end of the comparison statement.

```js
const isAvailable = useCallback(type => { 
   return paymentMethods.paymentMethods.length > 0 && paymentMethods.paymentMethods.some(pm => pm.type === type.toLowerCase());
}, [paymentMethods, paymentMethods.paymentMethods]);
```

Side note: It is a bit weird though to have `paymentMethods` object which has another `paymentMethods` property inside it 🤔

src/ComponentMap.js
```diff
@@ -0,0 +1,13 @@
...
+export function find(paymentMethods, type) {
+    return paymentMethods.paymentMethods.find(pm => pm.type === mapCreatedComponentType(type));
+}
+
+function mapCreatedComponentType(pmType) {
+    // Components created as 'card' need to be matched with paymentMethod response objects with type 'scheme'
+    return pmType === 'card' ? 'scheme' : pmType;
+}
+
+export const nativeComponents = [
```
ribeiroguilherme, line 11: Usually constants are uppercased. Most of devs follow [this rule](https://github.com/airbnb/javascript#naming--uppercase)

src/ComponentMap.js
```diff
@@ -0,0 +1,13 @@
+
```
ribeiroguilherme, line 1: It seems like this file is very similar to [this model](https://github.com/Adyen/adyen-web/blob/master/packages/lib/src/core/ProcessResponse/PaymentMethodsResponse/PaymentMethodsResponse.ts)  that we have in the Web repository. Creating proper models/classes for the Objects is a good idea, as it improves readability and encapsulates all the logic in one place.

This 'model' PaymentMethodsResponse could potentially be applied here as far as I see!

android/src/main/java/com/adyenreactnativesdk/component/AdyenInstantComponent.kt
```diff
@@ -0,0 +1,130 @@
...
+import org.json.JSONException
+import java.util.*
+
+class AdyenInstantComponent(context: ReactApplicationContext?) : BaseModule(context),
+    PaymentComponentListener, ActionHandlingInterface {
+
+    private var actionHandler: ActionHandler? = null
+
+    override fun getName(): String {
+        return "AdyenInstant"
```
OscarSpruit, line 25: I think it would be better to have this string as a static, so have a `private const val` in the `companion object`

ios/Components/InstantComponent.swift
```diff
@@ -0,0 +1,155 @@
...
+//
+
+
+import Adyen
+import Foundation
+import PassKit
+import React
+
+@objc(AdyenInstant)
+internal class InstantComponent: BaseModule {
```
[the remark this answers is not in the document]
descorp: 👍

ios/Components/InstantComponent.swift
```diff
@@ -0,0 +1,155 @@
...
+    
+    private var currentComponent: PresentableComponent?
+    private var currentPaymentComponent: PaymentComponent?
+    private var actionHandler: AdyenActionComponent?
+    
+    @objc
+    override static func requiresMainQueueSetup() -> Bool { true }
```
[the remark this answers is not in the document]
descorp: No, this is `override` of Obj-c code..

ios/Components/InstantComponent.swift
```diff
@@ -0,0 +1,155 @@
...
+    override func stopObserving() {}
+    override func startObserving() {}
+    override func supportedEvents() -> [String]! { super.supportedEvents() }
```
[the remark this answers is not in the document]
descorp: No, this is `override` of Obj-c code..

ios/Components/InstantComponent.swift
descorp: good point.. 😅

ios/Components/InstantComponent.swift
```diff
@@ -0,0 +1,155 @@
...
+
+    @objc
+    func hide(_ success: NSNumber, event: NSDictionary) {
+        DispatchQueue.main.async {
+            self.currentComponent?.finalizeIfNeeded(with: success.boolValue)
+            UIApplication.shared.keyWindow?.rootViewController?.dismiss(animated: true)
+
+            self.actionHandler = nil
+            self.currentComponent = nil
+            self.currentPaymentComponent = nil
```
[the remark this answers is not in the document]
descorp: 👍

ios/Components/InstantComponent.swift
descorp: Because we receive `NSDictionary` from RN but I prefere to parse it as `Decodable` - don't want to create new decoder for each class

ios/Components/InstantComponent.swift
descorp: Good point 🤔

ios/Components/InstantComponent.swift
descorp: 👍

src/ComponentMap.js
descorp: It is inspired by `PaymentMethodsResponse.ts`, yes :)
The only issue - there is no TypeScript yet applied to RN repo. 
I'll add this as improvement point for future!

src/ComponentMap.js
descorp: 👍

ios/Components/AdyenDropIn.swift
```diff
@@ -30,15 +30,28 @@ internal class AdyenDropIn: BaseModule {
...
 
     @objc
-    func open(_ paymentMethods: NSDictionary, configuration: NSDictionary) {
-        guard let data = try? JSONSerialization.data(withJSONObject: paymentMethods, options: []),
-              let paymentMethods = try? JSONDecoder().decode(PaymentMethods.self, from: data)
-        else { return }
+    func open(_ jsonData: NSDictionary, configuration: NSDictionary) {
+        let paymentMethods: PaymentMethods
+        
+        guard let data = try? JSONSerialization.data(withJSONObject: jsonData, options: []) else {
```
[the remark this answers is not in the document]
descorp: Because we receive `NSDictionary` from RN but I prefer to parse it as `Decodable` - don't want to create new decoder for each class

ios/Components/AdyenDropIn.swift
```diff
@@ -30,15 +30,28 @@ internal class AdyenDropIn: BaseModule {
...
+            return assertionFailure("AdyenDropIn: Can not deserialize paymentMethods.")
+        }
+        
+        if let jsonPaymentMethods = try? JSONDecoder().decode(PaymentMethods.self, from: data) {
+            paymentMethods = jsonPaymentMethods
+        }
+        else if let jsonPaymentMethod = try? JSONDecoder().decode(AnyPaymentMethod.self, from: data),
+                let anyPaymentMethod = jsonPaymentMethod.value {
```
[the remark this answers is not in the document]
descorp: Was working with both. But this code wasn't used..
Now it always expects a `PaymentMethods`

example/src/PaymentMethodsView.js
descorp: > It is a bit weird though to have paymentMethods object which has another paymentMethods property inside it 🤔

Good point. I'll rename it to `paymentMethodsResponse`.

android/src/main/java/com/adyenreactnativesdk/GenericPaymentComponentState.kt
```diff
@@ -0,0 +1,32 @@
...
+     */
+    val isReady: Boolean
+
+    /**
+     * @return If the collected data is valid to be sent to the backend.
+     */
+    val isValid: Boolean
+        get() = isInputValid && isReady
+
+    init {
```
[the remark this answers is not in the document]
descorp: 👍

android/src/main/java/com/adyenreactnativesdk/component/AdyenInstantComponent.kt
descorp: 👍

mohammedDehairy (approved):

src/ComponentMap.js
ribeiroguilherme: That is not an issue at all :-) Typescript helps to define types and make the syntax more strict and type-safe, but you can still create Models without TS by just using plain Javascript. You are at least able to encapsulate all the logic in that model. Later on, in case we migrate to TS, the refactoring will easier.

ribeiroguilherme (approved):

neelSharma12 (approved):

src/ComponentMap.js
descorp: Will try in next PR, thanks :)

Issue #53, opened 2022-10-24
Android AdyenInstant "stacking" `didProvide` calles


descorp:
```
To reproduce:
1. start payment on Android
2. cancel payment with the "cancel and go back to XX"
3. -> `handleDidProvide` gets called once with a cancel result
4. start payment again (in the same session)
5. -> `handleDidProvide` gets called twice with a cancel result
6. start payment again (in the same session)
7. -> `handleDidProvide` gets called three times with a cancel result
8. etc

The other (for us more serious) issue we're having that is also Android specific is:
1. start payment on Android
2. cancel payment with the "cancel and go back to XX"
3. -> `handleDidProvide` gets called once with a cancel result
4. start payment again (in the same session)
5. -> `handleDidProvide` gets called _immediately_ with a cancel result, without doing anything in the webview. This means that it is not possible to complete a payment once you've cancelled once because the cancel flow is triggered immediately.

```

_Originally posted by @kadikraman in https://github.com/Adyen/adyen-react-native/issues/44#issuecomment-1288802869_

descorp:
Hey @kadikraman 

This is "another bug", that I have [mentioned earlier](https://github.com/Adyen/adyen-react-native/issues/44#issuecomment-1278576565).
It happens due to the nature of Android ViewModel lifecycle: when PaymentComponent (any) attached to the Activity - it "contaminates" it with it's own state.
Fix for this will be an introduction of an extra fragment to isolate individual payment's lifecycle.
 
We are planing to fix for beta6 release as well ASAP

kadikraman:
Thank you @descorp - look forward to the fix!

descorp:
Hey @kadikraman 

This bug should be fixed in [1.0.0-beta6](https://github.com/Adyen/adyen-react-native/releases/tag/1.0.0-beta.6)
Could you verify?

kadikraman:
Can confirm, this is working for me now on beta 6 🎉

descorp:
Then I'll close this issue. Feel free to continue discussion here or reopen if problem occurs..

Issue #46, opened 2022-10-13
`onFail` callback does not get called on Android


kadikraman:
**Describe the bug**
If the payment is cancelled by closing the webview, the `onFail` does not get called on Android.

**To Reproduce**
1. Open the example app
2. Refetch payment methods
3. Start payment with a payment method that opens in a webview (e.g. PayPal or Klarna)
4. Close the webview
5. The `onFail` callback is not called on Android

**Expected behavior**
The `onFail` callback should be called so we could update the UI.

**Screenshots**
iOS - this is working as expected

https://user-images.githubusercontent.com/6534400/195580206-8ddd8040-d1e9-4f45-a86c-b2918f2c0d92.mp4

Android - this is not working

https://user-images.githubusercontent.com/6534400/195580243-2eb5b41f-9423-4699-92ac-21e0fba5bbcb.mp4

**Desktop (please complete the following information):**
 - OS: Android
 - Browser: Chrome
 - Version: 91.0.4472.114

**Smartphone (please complete the following information):**
 - Device:  Pixel 5
 - OS: API 32
 - Browser: Chrome
 - Version: 91.0.4472.114

faizantariq1:
i think for android your return url is not correct.

returnUrl: 'myapp://', // Custom URL scheme of your iOS app. This value is overridden for Android by `AdyenCheckout`. Can be send from your backend

kadikraman:
@faizantariq1 sorry, I don't quite understand: shouldn't the redirect work in the example app in this repo? I haven't changed any native code.

descorp:
Hey @kadikraman 

Thank you for a feedback! 
Indeed, going back on in-browser is buggy.

descorp:
@kadikraman 

I don't think there are any way to fix this (unless in current version of Android SDK).
For Example same issue could be observed on Drop-In or 3DS1 wed redirect 🤔

This is indeed very undesirables behaviour but Is this causing issues to you?

kadikraman:
Hi @descorp yes, this issue would be a release blocker for us, because once the user initialises the payment flow we do a bunch of things in the background (we convert a cart into an order, and show a loading state) - if the user cancels we'd need to undo these things in order for the user to be able to try a different payment method. However since we can't tell that the payment has failed, we don't know to trigger the cancel flow and the only way to try a different payment method would be to force close the app and start the checkout flow again.

[...]

PR #55, opened 2022-10-26
Add loading view to redirect action

descorp:
# Open PR

## Changes

* add cancelable loading view to redirect action

## Fixes

* #46 
* #53 

## Preview

ezgif-5-6b28c6ecf5

### DIFF
--- a/android/src/main/java/com/adyenreactnativesdk/ui/LoadingDialogFragment.kt
+++ b/android/src/main/java/com/adyenreactnativesdk/ui/LoadingDialogFragment.kt
@@ -1,0 +1,39 @@
+package com.adyenreactnativesdk.ui
+
+import android.graphics.Color
+import android.graphics.drawable.ColorDrawable
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import android.widget.Button
+import androidx.fragment.app.DialogFragment
+import com.adyenreactnativesdk.R
+
+interface Cancelable {
+    fun canceled()
+}
+
+class LoadingDialogFragment(private val cancelable: Cancelable) : DialogFragment() {
+
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
+        isCancelable = false
+        return inflater.inflate(R.layout.loading, container, false)
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        requireDialog().window?.setBackgroundDrawable(ColorDrawable(Color.TRANSPARENT))
+
+        val button = view.findViewById<Button>(R.id.loading_button)
+        button.setOnClickListener {
+            cancelable.canceled()
+        }
+    }
+
+    companion object {
+        fun newInstance(cancelable: Cancelable): LoadingDialogFragment {
+            return LoadingDialogFragment(cancelable)
+        }
+    }
+}
--- a/android/src/main/res/layout/loading.xml
+++ b/android/src/main/res/layout/loading.xml
@@ -5,28 +5,35 @@
     android:layout_width="match_parent"
     android:layout_height="match_parent">
 
-    <ProgressBar
-        android:id="@+id/progressBar"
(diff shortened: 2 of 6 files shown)


android/src/main/res/layout/loading.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent">
+
+    <ProgressBar
```
ozgur00, line 8: Maybe remove the progress bar so it's just an overlay with a button.

android/src/main/java/com/adyenreactnativesdk/ui/LoadingDialogFragment.kt
```diff
@@ -0,0 +1,39 @@
...
+
+        val button = view.findViewById<Button>(R.id.loading_button)
+        button.setOnClickListener {
+            cancelable.canceled()
+        }
+    }
+
+    companion object {
+        fun newInstance(cancelable: Cancelable): LoadingDialogFragment {
+            return LoadingDialogFragment(cancelable)
```
ozgur00, line 36: instead of passing the parameter in the constructor you should put it in arguments so your `newInstance()` 
```kotlin
val fragment = LoadingDialogFragment()
val args = Bundle().apply {
    putParcelable(KEY_CANCELLABLE, cancelable)
}
fragment.arguments = args
return fragment
}

but then you will have to make your `Cancelable` implementations `Parcelable`. you can use `@Parcelize` annotation for that. (you can take a look at this pr to see how to do that https://github.com/Adyen/adyen-android/pull/834)

File: android/src/main/java/com/adyenreactnativesdk/ui/LoadingDialogFragment.kt [hunk 3]
```diff
@@ -0,0 +1,39 @@
...
+import android.view.ViewGroup
+import android.widget.Button
+import androidx.fragment.app.DialogFragment
+import com.adyenreactnativesdk.R
+
+interface Cancelable {
+    fun canceled()
+}
+
+class LoadingDialogFragment(private val cancelable: Cancelable) : DialogFragment() {
```
ozgur00, line 17: Is the purpose gonna be loading? otherwise naming is confusing 😄

descorp: It is more of "Pending payment", indeed 🤔

android/src/main/java/com/adyenreactnativesdk/ui/LoadingDialogFragment.kt
```diff
@@ -0,0 +1,39 @@
...
+        requireDialog().window?.setBackgroundDrawable(ColorDrawable(Color.TRANSPARENT))
+
+        val button = view.findViewById<Button>(R.id.loading_button)
+        button.setOnClickListener {
+            cancelable.canceled()
+        }
+    }
+
+    companion object {
+        fun newInstance(cancelable: Cancelable): LoadingDialogFragment {
```
ozgur00, line 35: it's not a good practice to pass a callback in the `newInstance()` function and in the constructor. instead have a class level field as following:
```kotlin
var cancellable: Cancellable? = null
```

then set it to action handler where you instantiate the fragment.

android/src/main/res/layout/loading.xml
ozgur00: you can ignore my comment about this.

android/src/main/java/com/adyenreactnativesdk/ui/LoadingDialogFragment.kt
descorp: Fix here [38727d5](https://github.com/Adyen/adyen-react-native/pull/55/commits/38727d5a0b44c9205b5f9b2e133004a719e04ece)

ozgur00 (approved):

PR #56, opened 2022-10-26
Example: add navigation to Example App

descorp:
# Open PR

## Changes

* add Result Page
* add Home Page

## Preview

ezgif-2-26b3a96ff6

### DIFF
--- a/example/src/CheckoutView.js
+++ b/example/src/CheckoutView.js
@@ -1,8 +1,7 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import { AdyenCheckout } from '@adyen/react-native';
 import { fetchPayments, fetchPaymentDetails, isSuccess } from './APIClient';
 import {
-  Button,
   SafeAreaView,
   StatusBar,
   StyleSheet,
@@ -12,7 +11,7 @@ import {
   Alert,
 } from 'react-native';
 import { Colors } from 'react-native/Libraries/NewAppScreen';
-import { PaymentMethodsContext } from './PaymentMethodsProvider';
+import { usePaymentMethods } from './PaymentMethodsProvider';
 import PaymentMethods from './PaymentMethodsView';
 import { ERROR_CODE_CANCELED } from '../../src';
 
@@ -22,7 +21,7 @@ const styles = StyleSheet.create({
     borderRadius: 5,
     justifyContent: 'center',
     padding: 16,
-  },
+  }
 });
 
 function getFlagEmoji(countryCode) {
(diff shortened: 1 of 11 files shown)


example/src/CheckoutView.js
```diff
@@ -86,42 +86,50 @@ const CheckoutView = () => {
     }
   };
 
+  const { config, paymentMethods, onConfigChanged } = usePaymentMethods();
```
ribeiroguilherme, line 89: It is good practice to group all `useXXX` hooks in the beginning of the function Component - I'd suggest to put this one right after the line 36 :)  

This way it is clear from the beginning of the code from where this Component is using resources.

example/src/CheckoutView.js
```diff
@@ -86,42 +86,50 @@ const CheckoutView = () => {
...
+
+  useEffect(() => { onConfigChanged(config) },
+    [paymentMethods == null],
+  );
```
ribeiroguilherme, line 93: This `useEffect` doesn't seem right 🤔  

In the dependency array you usually pass object instances or primitives, and in case those object references get updated, then the effect is triggered. In this current scenario, we are doing a direct comparison check here.

What is your end goal here? What is the use-case of this use effect?

example/src/CheckoutView.js
```diff
@@ -86,42 +86,50 @@ const CheckoutView = () => {
...
+        onFail={didFail}
+        onComplete={didComplete}
+      >
+        <PaymentMethods />
+      </AdyenCheckout>
+    </SafeAreaView>
   );
 };
 
+const PaymentMethodsView = (props) => {
```
ribeiroguilherme, line 115: You can destructure the props here, so the Component code would be less verbose:

```jsx
const PaymentMethodsView = ({ paymentMethods, config }) => { ...
```

example/src/CheckoutView.js
descorp: 👍

example/src/CheckoutView.js
descorp: 👍

descorp: Fixed in [17aae14](https://github.com/Adyen/adyen-react-native/pull/56/commits/17aae14ac8766f76a157d7198033915d4d4426e6)

example/src/CheckoutView.js
descorp: Fixed in [17aae14](https://github.com/Adyen/adyen-react-native/pull/56/commits/17aae14ac8766f76a157d7198033915d4d4426e6)

example/src/CheckoutView.js
descorp: Fixed here [9608f81](https://github.com/Adyen/adyen-react-native/pull/56/commits/9608f811e255b756aee1474882d00fc2893ca9f4)

ribeiroguilherme (approved):
So pretty now! 🥇

PR #105, opened 2022-12-16
Add support for new type 'googlepay'

descorp:
# Open PR

### Fix

* add support for `googlepay` payment method type. Both `googlepay` and `paywithgoogle` are supported

### DIFF
--- a/android/src/main/java/com/adyenreactnativesdk/component/BaseModule.kt
+++ b/android/src/main/java/com/adyenreactnativesdk/component/BaseModule.kt
@@ -10,9 +10,7 @@ import android.os.Build
 import androidx.appcompat.app.AppCompatActivity
 import com.adyen.checkout.components.model.PaymentMethodsApiResponse
 import com.adyen.checkout.components.model.paymentmethods.PaymentMethod
-import com.adyen.checkout.components.model.payments.response.Action
 import com.adyenreactnativesdk.action.ActionHandler
-import com.adyenreactnativesdk.configuration.RootConfigurationParser
 import com.adyenreactnativesdk.util.ReactNativeError
 import com.adyenreactnativesdk.util.ReactNativeJson
 import com.facebook.react.bridge.ReactApplicationContext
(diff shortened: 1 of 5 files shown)


android/src/main/java/com/adyenreactnativesdk/component/BaseModule.kt
```diff
@@ -48,6 +48,17 @@ abstract class BaseModule(context: ReactApplicationContext?) : ReactContextBaseJ
         }
     }
 
+    protected fun getPaymentMethod(
+        paymentMethodsResponse: PaymentMethodsApiResponse,
+        paymentMethodNames: Set<String>
+    ): PaymentMethod? {
+        for (currentPaymentMethod in paymentMethodsResponse.paymentMethods ?: emptyList())
```
ozgur00: This can be simplified by 
```kotlin
return paymentMethodsResponse.paymentMethods?.firstOrNull { paymentMethodNames.contains(it.type) }
```

example/src/PaymentMethodsView.js
```diff
@@ -27,7 +27,7 @@ const PaymentMethods = () => {
   const platformSpecificPayment =
     Platform.OS === 'ios' ? 'Apple Pay' : 'Google Pay';
   const platformSpecificType =
-    Platform.OS === 'ios' ? 'applepay' : 'paywithgoogle';
+    Platform.OS === 'ios' ? 'applepay' : 'googlepay';
```
ozgur00: Other places I see we kept `paywithgoogle` but not here. I don't know what this line does so not 100% sure how removing `paywithgoogle` will affect here.

src/AdyenNativeModules.js
```diff
@@ -50,6 +50,8 @@ export function getNativeComponent(name, paymentMethods) {
     case 'apple-pay':
       return { nativeComponent: AdyenApplePay };
     case 'paywithgoogle':
+    case 'googlepay':
+    case 'google-pay':
```
ozgur00: Why do we need `google-pay`? I see we did the same for apple pay.

example/src/PaymentMethodsView.js
descorp: This is example app, but it could potentially cause issue and frustration to some merchant. 
Added comment: [9b8737b](https://github.com/Adyen/adyen-react-native/pull/105/commits/9b8737ba6198192e2c97dcdcba07765a7954eab7)

src/AdyenNativeModules.js
descorp: Not really needed. I'll remove it 🤔

descorp: fixed [14cfd19](https://github.com/Adyen/adyen-react-native/pull/105/commits/14cfd19033353fdc4fea058ac90564e7e5433d4b)

android/src/main/java/com/adyenreactnativesdk/component/BaseModule.kt
descorp: fixed [6788eb9](https://github.com/Adyen/adyen-react-native/pull/105/commits/6788eb96b96f27e0ddb86d2328a52f4d95d5800d)

ozgur00 (approved):

PR #147, opened 2023-02-22
Add Client side Encryption

descorp:
# Open PR

### New

* add  native module AdyenCSE with methods 
  * `encryptCard(card: Object, publicKey: String) -> Promise`
  * `encryptBin(bin: String, publicKey: String) -> Promise`

### DIFF
--- a/android/src/main/java/com/adyenreactnativesdk/cse/AdyenCSEModule.kt
+++ b/android/src/main/java/com/adyenreactnativesdk/cse/AdyenCSEModule.kt
@@ -1,0 +1,33 @@
+package com.adyenreactnativesdk.cse
+
+import com.facebook.react.bridge.ReactApplicationContext
+import com.facebook.react.bridge.ReactContextBaseJavaModule
+import com.facebook.react.bridge.ReactMethod
+import com.facebook.react.bridge.ReadableMap
+
+class AdyenCSEModule(context: ReactApplicationContext?) : ReactContextBaseJavaModule(context) {
+
+    @ReactMethod
+    fun addListener(eventName: String?) { }
+
+    @ReactMethod
+    fun removeListeners(count: Int?) { }
+    
+    override fun getName(): String {
+        return COMPONENT_NAME
+    }
+
+    @ReactMethod
+    fun encryptCard(card: ReadableMap, publicKey: String) {
+    }
+
+
+    @ReactMethod
+    fun encryptBin(bin: String, publicKey: String) {
+    }
+
+    companion object {
+        private const val TAG = "AdyenCSE"
+        private const val COMPONENT_NAME = "AdyenCSE"
+    }
+}
--- a/android/src/main/java/com/adyenreactnativesdk/cse/AdyenCSEModule.kt
+++ b/android/src/main/java/com/adyenreactnativesdk/cse/AdyenCSEModule.kt
@@ -32,7 +32,7 @@ class AdyenCSEModule(context: ReactApplicationContext?) : ReactContextBaseJavaMo
             map.putString(CVV_KEY, encryptedCard.encryptedSecurityCode)
             promise.resolve(map)
         } catch (e: EncryptionException) {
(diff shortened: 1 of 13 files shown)


android/src/main/java/com/adyenreactnativesdk/cse/AdyenCSEModule.kt
```diff
@@ -0,0 +1,58 @@
...
+            map.putString(NUMBER_KEY, encryptedCard.encryptedCardNumber)
+            map.putString(EXPIRY_MONTH_KEY, encryptedCard.encryptedExpiryMonth)
+            map.putString(EXPIRY_YEAR_KEY, encryptedCard.encryptedExpiryYear)
+            map.putString(CVV_KEY, encryptedCard.encryptedSecurityCode)
+            promise.resolve(map)
+        } catch (e: EncryptionException) {
+            promise.reject(ERROR_MESSAGE, e)
+        }
+    }
+
```
OscarSpruit, line 38: The `CardEncrypter` also has the `public static String encrypt(...)` method to encrypt all data into a single string. Did you leave that out on purpose?

descorp: AFAIK it is for "Classic" PAL integration that was deprecated. Unless it is requested by someone - I see no reason adding it.

OscarSpruit (approved):

erenbesel (approved):

neelSharma12 (approved):

example/src/CSE.js
```diff
@@ -0,0 +1,50 @@
...
+    try {
+      const encryptedCard = await AdyenCSE.encryptCard(
+        unencryptedCard,
+        publicKey
+      );
+      console.log(JSON.stringify(encryptedCard));
+    } catch (e) {
+      console.error(e);
+    }
+  };
```
ribeiroguilherme, line 27: You can wrap the declared functions in `useCallback` hook - this way, they aren't recreated in memory if the Component re-renders.

```jsx
const tryEncryptCard = useCallback(async () => {
...
}. []);
```

example/src/CSE.js
```diff
@@ -0,0 +1,50 @@
...
+import { Button, SafeAreaView } from 'react-native';
+import { NativeModules } from 'react-native';
+import { ENVIRONMENT } from './Configuration';
+import Styles from './Styles';
+
+const { AdyenCSE } = NativeModules;
+
+const CseView = () => {
+  const tryEncryptCard = async () => {
+    let unencryptedCard = {
```
ribeiroguilherme, line 11: Should be defined as `const` instead of `let`, as there is no need to mutate the object

example/src/CSE.js
```diff
@@ -0,0 +1,50 @@
...
+      number: '5454 5454 5454 5454',
+      expiryMonth: '03',
+      expiryYear: '2030',
+      cvv: '737',
+    };
+    let publicKey = ENVIRONMENT.publicKey;
```
ribeiroguilherme, line 17: Perhaps this publicKey could be moved to the top of the file after the imports, as it is a constant value used in this Component. You don't have to define it locally in both functions

```jsx
...
import { ENVIRONMENT } from './Configuration';
import Styles from './Styles';

const PUBLIC_KEY = ENVIRONMENT.publicKey;
...
```

ribeiroguilherme (approved):

example/src/CSE.js
descorp: This will probably change (or maybe even removed).
I'll address this in next PR

example/src/CSE.js
descorp: This will probably change (or maybe even removed).
I'll address this in next PR

example/src/CSE.js
descorp: This will probably change (or maybe even removed).
I'll address this in next PR

PR #152, opened 2023-03-03
Refactor example app

descorp:
# Open PR

### New

* Update structure of Example app
* Apply Styles across Example app
* Add custom card sample to CseView


example/src/Views/Checkout/PaymentMethodsView.js
```diff
@@ -0,0 +1,86 @@
...
+  const platformSpecificPayment =
+    Platform.OS === 'ios' ? 'Apple Pay' : 'Google Pay';
+  const platformSpecificType = Platform.OS === 'ios' ? 'applepay' : 'googlepay'; // In some cases 'paywithgoogle' can be in use. Check paymentMethods response first.
+
+  const isAvailable = useCallback(
+    (type) => {
+      if (!paymentMethodsResponse) {
+        return false;
+      }
+      const paymentMethods = paymentMethodsResponse.paymentMethods;
```
ribeiroguilherme:
```suggestion
      const { paymentMethods } = paymentMethodsResponse;
```

example/src/Views/Checkout/PaymentMethodsView.js
```diff
@@ -0,0 +1,86 @@
...
+        paymentMethods.length > 0 &&
+        paymentMethods.some(
+          (pm) => pm.type.toLowerCase() === type.toLowerCase()
+        )
+      );
+    },
+    [paymentMethodsResponse]
+  );
+
+  const isReady = paymentMethodsResponse === null;
```
ribeiroguilherme: `isReady` when paymentMethodsResponse is null? 🤔  Sounds weird.. shouldn't be the opposite? Seems like the variable name is incorrect

example/src/Views/CseView.js
```diff
@@ -0,0 +1,105 @@
...
+const PUBLIC_KEY = ENVIRONMENT.publicKey;
+
+const CseView = () => {
+  const [number, setNumber] = useState('');
+  const [expiryMonth, setExpiryMonth] = useState('');
+  const [expiryYear, setExpiryYear] = useState('');
+  const [cvv, setCvv] = useState('');
+
+  const tryEncryptCard = useCallback(async () => {
+    let unencryptedCard = {
```
ribeiroguilherme:
```suggestion
    const unencryptedCard = {
```

example/src/Views/CseView.js
```diff
@@ -0,0 +1,105 @@
...
+      expiryMonth: expiryMonth,
+      expiryYear: expiryYear,
+      cvv: cvv,
+    };
+    try {
+      const encryptedCard = await AdyenCSE.encryptCard(
+        unencryptedCard,
+        PUBLIC_KEY
+      );
+      let data = {
```
ribeiroguilherme:
```suggestion
      const data = {
```

example/src/Views/CseView.js
```diff
@@ -0,0 +1,105 @@
...
+        paymentMethod: {
+          type: 'scheme',
+          encryptedCardNumber: encryptedCard.number,
+          encryptedExpiryMonth: encryptedCard.expiryMonth,
+          encryptedExpiryYear: encryptedCard.expiryYear,
+          encryptedSecurityCode: encryptedCard.cvv,
+        },
+      };
+      let result = await fetchPayments(data, DEFAULT_CONFIGURATION);
```
ribeiroguilherme:
```suggestion
      const result = await fetchPayments(data, DEFAULT_CONFIGURATION);
```

src/AdyenNativeModules.js
```diff
@@ -55,7 +82,7 @@ export function getNativeComponent(name, paymentMethods) {
       break;
   }
 
-  let paymentMethod = find(paymentMethods, type)
+  let paymentMethod = find(paymentMethods, type);
```
ribeiroguilherme:
```suggestion
  const paymentMethod = find(paymentMethods, type);
```

ribeiroguilherme (dismissed):

example/src/Views/Checkout/PaymentMethodsView.js
descorp: Good point 😅

descorp: Fixed [ecc48f4](https://github.com/Adyen/adyen-react-native/pull/152/commits/ecc48f44af112d14c07f78aa61873cec3739a511)

example/src/Utilities/Styles.js
```diff
@@ -0,0 +1,36 @@
+import { StyleSheet } from 'react-native';
+
+const Styles = StyleSheet.create({
+  page: {
+    flex: 1,
+  },
+  content: {
+    alignItems: 'center',
+    flex: 1,
```
m1aw, line 9: Are you sure `flex: 1` is what you want? This is just the shorthand for flex-grow/flex-shrink/flex-basis https://developer.mozilla.org/en-US/docs/Web/CSS/flex

To enable "flex" features use `display: flex` https://developer.mozilla.org/en-US/docs/Web/CSS/display

descorp: All the React Native container elements are, by default, Flex containers. There’s no need to add a display: ‘flex’.

In this case `flex: 1` is for "page" to take all available space so child item "content" item can position in a center of the screen.

ribeiroguilherme (approved):

PR #155, opened 2023-03-08
JsDocs update

descorp:
# Open PR

### New

* provide JsDocs to `AdyenCheckout` props
* correct meta-docs and remove constrints
* rename `onProvide` to `onAdditionalDetails`
* rename `onFail` to `onError`

### Fixes

* typo on iOS
* typo in

### DIFF
--- a/src/AdyenCheckoutContext.js
+++ b/src/AdyenCheckoutContext.js
@@ -69,8 +69,8 @@ const AdyenCheckout = (props) => {
   const subscriptions = useRef([]);
   const { config, paymentMethods, onSubmit, onComplete, children } = props;
 
-  let onError = props.onError == undefined ? props.onFail : props.onError;
-  let onAdditionalDetails =
+  const onError = props.onError == undefined ? props.onFail : props.onError;
+  const onAdditionalDetails =
     props.onAdditionalDetails == undefined
       ? props.onProvide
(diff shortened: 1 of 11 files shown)


src/AdyenCheckoutContext.js
```diff
@@ -1,24 +1,79 @@
...
+ */
+
+/** @param {AdyenCheckoutProp} props */
+const AdyenCheckout = (props) => {
+  /** @type {import('react').MutableRefObject<import('react-native').EventSubscription[]>} */
   const subscriptions = useRef([]);
+  const { config, paymentMethods, onSubmit, onComplete, children } = props;
+
+  let onError = props.onError == undefined ? props.onFail : props.onError;
+  let onAdditionalDetails =
```
ribeiroguilherme: Always use `const` instead of `let`. If the compiler complain, then you fix it. Most (if not all) the cases you are going to use `const`

ribeiroguilherme (dismissed):

src/AdyenCheckoutContext.js
descorp: Fixed in [30f2220](https://github.com/Adyen/adyen-react-native/pull/155/commits/30f2220b7d503574eb14366aee50dcc192488eae)

ribeiroguilherme (approved):

PR #156, opened 2023-03-09
Example refactor

descorp:
# Open PR

## Changes

* Example App: replace static methods with ApiClient class
* improve NPM package description

### DIFF
--- a/example/src/Utilities/PaymentMethodsProvider.js
+++ b/example/src/Utilities/PaymentMethodsProvider.js
@@ -15,32 +15,28 @@ export const usePaymentMethods = () => {
 };
 
 const PaymentMethodsProvider = (props) => {
-  const [state, setState] = useState({
-    config: DEFAULT_CONFIGURATION,
-    paymentMethods: undefined,
-  });
+  const [config, setConfig] = useState(DEFAULT_CONFIGURATION);
+  const [paymentMethods, setPaymentMethods] = useState(undefined);
 
   const refresh = useCallback(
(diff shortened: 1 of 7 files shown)


example/src/Utilities/PaymentMethodsProvider.js
```diff
@@ -14,36 +14,39 @@ export const usePaymentMethods = () => {
...
 
-class PaymentMethodsProvider extends Component {
-  state = {
+const PaymentMethodsProvider = (props) => {
+  const [state, setState] = useState({
     config: DEFAULT_CONFIGURATION,
-    paymentMethods: null,
-  };
+    paymentMethods: undefined,
+  });
```
ribeiroguilherme: There is nothing wrong doing this, but the rule of thumb is that `useState` ideally should manipulate only one type of object. There is no problem creating multiple states.

In this case, it is better to:
```jsx
const [config, setConfig] = useState(DEFAULT_CONFIGURATION);
const [paymentMethods, setPaymentMethods]= useState(undefined);
```

ribeiroguilherme (dismissed):

example/src/Utilities/PaymentMethodsProvider.js
descorp: Fixed on [c5caa73](https://github.com/Adyen/adyen-react-native/pull/156/commits/c5caa7340d01274755a3e945975441296dcdb11e)

ribeiroguilherme (approved):

PR #166, opened 2023-03-30
Type script

descorp:
# Open PR

## Changes

* rewrite everything on Typescript
* add type checks
* add configuration checks
* change constants names to lower-case


src/Core/constants.ts
```diff
@@ -0,0 +1,85 @@
...
+  noPaymentMethod = 'noPaymentMethod',
+}
+
+/** Collection of available environments. */
+export enum Environment {
+  test = 'test',
+  europe = 'live-eu',
+  us = 'live-us',
+  australia = 'live-au',
+  asiaPacificSouthEast = 'live-apse',
```
m1aw: I have an opinion, so feel free to ignore it and proceed with this change.
I think instead of some of these enums, String literal are better since it keeps it consistent with the documentation that already exists.
https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-types

m1aw (dismissed):

src/Core/constants.ts
descorp: Fixed in [e3b74f5](https://github.com/Adyen/adyen-react-native/pull/166/commits/e3b74f591147b1df0005956f36fc28c3fe6222ab)

m1aw (dismissed):

tsconfig.json
```diff
@@ -1,17 +1,18 @@
 {
-  "include": [ "src/**/*" ],
+  "include": ["./src/**/*"],
+  "exclude": ["./lib/**/*"],
```
ribeiroguilherme, line 3: What is this `lib` folder? It doesn't seem to exist

src/index.ts
```diff
@@ -1,30 +1,40 @@
...
-export { Environment, Event, ErrorCode, ResultCode } from './constants';
+export {
+  Configuration,
+  DropInConfiguration,
+  CardAuthMethod,
+  ApplePayConfiguration,
+  GooglePayConfiguration,
+  AddressMode,
+  FieldVisibility,
+} from './Core/configuration';
```
ribeiroguilherme, line 36: Did you try by any chance to integrate these changes within a React Native app written in Typescript?  Ex: build the library files, and then import it within other app 

That can be a good way to test if the types are working as expected . For now, your code is typed and it works fine on your IDE and when developing, but it is important to double-check if it also works well when other Typescript app start to integrate with it.

ribeiroguilherme (dismissed):

tsconfig.json
descorp: It is folder for generated binary. Could be also called 'build' 🤔

src/AdyenNativeModules.tsx
```diff
@@ -0,0 +1,170 @@
...
+  PaymentMethodsResponse,
+} from './Core/types';
+
+/** Universal interface for Adyen Native payment component */
+interface AdyenComponent {
+  /** Show component above current screen. */
+  open: (paymentMethods: PaymentMethodsResponse, configuration: any) => void;
+
+  /** Dismiss component from screen. */
+  hide: (success: boolean, message: string | undefined) => void;
```
gkueny: As we can see in exemple

```ts
nativeComponent.hide(true, { message: 'Completed' });
```

shouldn't the type be something like this ?

```suggestion
  hide: (success: boolean, event?: { message?: string }) => void;
```

gkueny: With latest version on `develop` branch

`nativeComponent.hide(true, 'Completed');`  failed with Swift error on iOS, while `nativeComponent.hide(true,  { message: 'Completed' });` succeed

descorp: Hey @gkueny 

Thanks for feedback! 💚
This is good find.

descorp: Fixed in [0a94c14](https://github.com/Adyen/adyen-react-native/pull/166/commits/0a94c14ce72c92e340cc2deb59a6b67c2751a636)

src/index.ts
descorp: Yes. Checked on TS Demo project 😎

ribeiroguilherme (approved):

Issue #175, opened 2023-04-12
Bad type for `google pay` configuration


gkueny:
## Bug

In the definition of type `Configuration`, `google` key is used instead of `googlepay`.

## Reproduction

Setup GooglePay config with `google` key.
An error "GooglePay merchantAccount not found" is throw.

## Expected behavior

With `googlepay` key, payment work as expected

descorp:
> An error "GooglePay merchantAccount not found" is throw.

In general `merchantAccount` is passed in [\paymentMethods](https://docs.adyen.com/api-explorer/Checkout/68/post/paymentMethods); setting it manually (unlike for Apple Pay) is not required

gkueny:
yeah, sadly our backend team stripe this info from `paymentMethods` so we set it manually.

But with `googlepay` key, everything work fine for us !

PR #176, opened 2023-04-12
:bug: (googlepay) fix googlepay configuration key type

gkueny:
## Summary

In the definition of type Configuration, `google` key is used instead of `googlepay`.

## Tested scenarios

- Payement with googlepay 

**Fixed issue**:  #175


descorp (approved):

descorp:
Thanks again @gkueny 💚

this one is super nasty!

PR #187, opened 2023-04-13
example typescript support

pinpong:
Update to latest react-native template because i was not able to run the android example.


pinpong:
@descorp due you latest changes the PR needs to be resolved the conflicts. Is it worth or not?

descorp:
Sorry @pinpong 
Had no time to check your PR..

First of all, thanks for your contribution! 💚
Your PR has multiple nice additions that we want to cherry-pick.  

We are grateful for your time and we don't want to waist any of it :)
Some concerns we still have about this PR:

> latest react-native template

Updating to latest is never-ending chaise.
Current example app is kept as "legacy compatibility reference" with JS and "old architecture".

We also have pre-release testing strategy: we publishing local NPM and running smoke tests on blank TS project and "new architecture"; In future we will attempt to automate this with CI. 

This approach expected to provide backward and forward compatibility.

> example typescript support

For now we keeping example app in JS+JS-docs so it is easy to "copy-paste" for everyone. 
Rational: JS-only consumers will be able to copy it "as-is" and TS consumers can provide minor modifications. If whole project will be in TS - it would be take more effort for JS-only consumers.

---

Feel free to share your opinion, we are looking for feedback of actual consumers who worked with our SDK 🙂

example/android/app/src/main/java/com/adyenexample/MainActivity.java
```diff
@@ -24,7 +28,7 @@ protected String getMainComponentName() {
 
   @Override
   protected void onCreate(Bundle savedInstanceState) {
-    super.onCreate(savedInstanceState);
+    super.onCreate(null);
```
descorp, line 31: I am wondering why is this necessary?

example/src/Views/Checkout/PaymentMethodsView.tsx
```diff
@@ -0,0 +1,64 @@
...
+            <View key={`${p.type}`} style={Styles.item}>
+              <Button
+                title={`${p.type}`}
+                disabled={
+                  isNotReady ||
+                  !isAvailable(p.type) ||
+                  (Platform.OS !== 'ios' && p.type === 'applepay') ||
+                  (Platform.OS !== 'android' && p.type === 'googlepay') ||
+                  /// In some cases 'paywithgoogle' can be in use. Check paymentMethods response first.
+                  (Platform.OS !== 'android' && p.type === 'paywithgoogle')
```
descorp:
```suggestion
                  (Platform.OS !== 'android' && ( p.type === 'googlepay' || p.type === 'paywithgoogle' ))
                  /// In some cases 'paywithgoogle' can be in use. Check paymentMethods response first.
```

example/android/app/src/main/java/com/adyenexample/MainActivity.java
pinpong: You can check here: https://reactnavigation.org/docs/7.x/getting-started
and more detailed here: https://github.com/software-mansion/react-native-screens/issues/17

pinpong:
Agree with the  never-ending story about upgrading react.native. I personally was surprised that the example was written in JS and it took me more time to understand it.

PR #193, opened 2023-04-25
fixed payments

pinpong:
fixed payments e.g. klarna

### DIFF
--- a/example/src/Utilities/APIClient.js
+++ b/example/src/Utilities/APIClient.js
@@ -11,6 +11,7 @@ class ApiClient {
     const body = {
       ...data,
       ...parseConfig(configuration),
+      ...parseAmount(configuration),
       ...serverConfiguration,
       additionalData: { allow3DS2: true },
       lineItems: [
--- a/example/src/Utilities/APIClient.js
+++ b/example/src/Utilities/APIClient.js
@@ -1,7 +1,7 @@
 // @ts-check
 
 import { LogBox } from 'react-native';
-import { ENVIRONMENT, CHANNEL, DEVICE_LOCALE } from '../Configuration';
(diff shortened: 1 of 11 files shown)


example/src/Utilities/APIClient.js
```diff
@@ -11,6 +11,7 @@ class ApiClient {
     const body = {
       ...data,
       ...parseConfig(configuration),
+      ...parseAmount(configuration),
```
descorp: Thanks for pointing this problem out!
This is interesting one:

For `\payments` body, the `amount` is **expected** to be taken from  `data`.
However, If you do not provide `amount` in `AdyenContext` configuration - you are going to get API error.

I would suggest to do something like: 
```js
const parseAmount = ({ amount, currency }, data) => ({
  amount: data.amount ?? { value: amount, currency: currency },
});
```

---

Some context:

`amount` is **required** for [`\payment` API endpoint](https://docs.adyen.com/api-explorer/Checkout/68/post/payments).

The rule of thumb: that merchant's backend is a final source of truth for `amount`. For obvious reasons merchant should never trust an `amount` value coming from client side.

It is essential that shopper pays exact amount they agree to pay. 
For this reason our `AdyenContext` configuration  have `amount` property: it allow to show "Pay ### $$$" on payment component UI; this `amount` passed thought component to `paymentData` object you receive on `onSubmit`.

> ❗  It is a good practice to set `amount` in configuration.

Side note:
In specific cases (giftcards, mealvoucher, instalments etc), amount may change on client side. 
In this case it is especially important to verify `amount` coming from component/Drop-In.

descorp:
Is this working for both scenarios:

a) `amount` set to `AdyenContext` configuration
b) `amount` not set to `AdyenContext` configuration

PR #198, opened 2023-04-26
dismiss dialog on hot reload

pinpong:
Improved developer experience by dismissing the pending dialog fragment on hot reload.

How to proof:
1. Open the example app and open e.g. klarna
2. close the webview
3. hot reload
4. the dialog is no closable because the js bridge was destroyed


descorp:
Hey @pinpong

Thanks for contribution 💚

Same as #201:
This is a nice addition from a development point of view, however,  need to do more testing to make sure this can't mess up with Payments flow state in some way 🤔

descorp:
Since `onCatalystInstanceDestroy` was deprecated - I'll close this PR.
Thanks for contribution @pinpong

PR #200, opened 2023-04-27
fixed ios build on xcode 14.3

pinpong:
react-native >= [0.70.8](https://github.com/facebook/react-native/releases/tag/v0.70.8) required to build with xcode 14.3


descorp:
Our iOS SDK officially targeting version is **11.0** :)
We are looking to move React Native to iOS **13.0** with v5 update (around July).

dJani97:
I also had to set this manually everywhere when running the example project. It would be nice if it would be the default.

descorp:
I am going to close this one for now.
Example app was updated to [0.71.9](https://github.com/facebook/react-native/releases/tag/v0.71.9) and main lib don't have direct dependency.

Thanks @pinpong !

PR #201, opened 2023-04-27
fixed hot realod

pinpong:
Improved developer experience by hide the payment components.

How to proof:

Open the example app and dropin
hot reload
action sheet is not closable because the js bridge was destroyed


descorp:
Hey @pinpong 

Thanks for contribution 💚

This is a nice addition from a development point of view - no need to kill and restart app after reloading ReactNative debugger.
However, I am still doubtful if this can mess up with Payments flow state in some way 🤔

Need to do more testing..

descorp:
Thanks @pinpong 💚

We have recreated same logic for **DEBUG** only in #251

PR #202, opened 2023-04-27
fix configuration

pinpong:
This PR fixed the configuration issue on the example.
It also extends the type for configuration.

@descorp when type script example ;D

### DIFF
--- a/example/src/Configuration.js
+++ b/example/src/Configuration.js
@@ -12,22 +12,54 @@ export const CHANNEL = Platform.select({
 })();
 
 export const DEFAULT_CONFIGURATION = {
-  countryCode: 'NL',
-  currency: 'EUR',
-  amount: 1000, // The amount value in minor units.
+  environment: 'test',
   merchantAccount: '{YOUR_MERCHANT_ACCOUNT}',
   merchantName: 'MyStore',
-  applepayMerchantID: '{YOUR_APPLE_MERCHANT_ID}',
-  shopperLocale: DEVICE_LOCALE,
+  clientKey: '{YOUR_CLIENT_KEY}',
+  returnUrl: 'myapp://payment', // Only used for iOS,
   shopperReference: 'Checkout Shopper',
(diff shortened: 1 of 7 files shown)


src/Core/configuration.ts
```diff
@@ -16,6 +16,12 @@ export interface Configuration {
   /** Selected environment */
   environment: Environment;
 
+  /** Adyen merchant account. */
+  merchantAccount: string;
+
+  /** Merchant name to be displayed on payment summary item. */
+  merchantName: string;
```
descorp: This fields are not part of `AdyenContext` Configuration:

* `merchantAccount` is only used for API calls. It is exist in scope of "Example App" because it makes requests to backend, but **real live** Adyen integration should not have this property on client side.
* `merchantName` is used for ApplePay only.

src/Core/configuration.ts
```diff
@@ -25,6 +31,9 @@ export interface Configuration {
   /** The shopper's country code. A valid value is an ISO two-character country code (e.g. 'NL'). */
   countryCode?: string;
 
+  /** The shopper's locale code. A valid value is an ISO locale code (e.g. 'nl-NL'). */
+  shopperLocale: string;
```
descorp, line 30: This property is not yet supported on iOS, hence was abandoned for now..

example/src/Configuration.js
```diff
@@ -12,22 +12,54 @@ export const CHANNEL = Platform.select({
 })();
 
 export const DEFAULT_CONFIGURATION = {
```
descorp, line 14: The idea of `DEFAULT_CONFIGURATION` is to provide a plain list of constants and properties and feature-toggles for App's "Configuration".
All necessary payment configurations are expected on `AdyenCheckout` directly - this will act as code sample, so it is easy to copy-past

descorp:
Hey @pinpong 

I have split this PR into two #222 and #223

Thank you for your contribution! 💚

PR #204, opened 2023-04-28
fixed requires main queue setup warning

pinpong:
Fixed:
```console
 WARN  Module AdyenCSE requires main queue setup since it overrides `init` but doesn't implement `requiresMainQueueSetup`. In a future release React Native will default to initializing all native modules on a background thread unless explicitly opted-out of.
```


descorp (approved):

descorp:
Thanks @pinpong 💚

PR #210, opened 2023-05-05
Include card brand to JSON on Submit

descorp:
# Open PR

## Changes

* now for Android `brand` is included to JSON object on Submit

### DIFF
--- a/android/src/main/java/com/adyenreactnativesdk/component/dropin/AdyenCheckoutService.kt
+++ b/android/src/main/java/com/adyenreactnativesdk/component/dropin/AdyenCheckoutService.kt
@@ -6,6 +6,7 @@
 package com.adyenreactnativesdk.component.dropin
 
 import android.util.Log
+import com.adyen.checkout.card.CardComponentState
 import com.adyen.checkout.components.ActionComponentData
 import com.adyen.checkout.components.PaymentComponentState
 import com.adyen.checkout.components.model.payments.response.Action
@@ -14,11 +15,6 @@ import com.adyen.checkout.dropin.service.DropInServiceResult
 import com.adyenreactnativesdk.component.dropin.DropInServiceProxy.DropInModuleListener
 import com.facebook.react.bridge.ReadableMap
 import org.json.JSONObject
-
-@Deprecated(
-    message = "This class is deprecated on beta-9",
-    replaceWith = ReplaceWith("AdyenCheckoutService"))
-class AdyenDropInService: AdyenCheckoutService() { }
 
 open class AdyenCheckoutService : DropInService(), DropInModuleListener {
     override fun onCreate() {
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/adyenreactnativesdk/component/dropin/AdyenCheckoutService.kt
```diff
@@ -31,6 +27,11 @@ open class AdyenCheckoutService : DropInService(), DropInModuleListener {
         paymentComponentJson: JSONObject
     ) {
         val listener = DropInServiceProxy.shared.serviceListener
+        if (paymentComponentState is CardComponentState && paymentComponentState.cardType != null) {
```
jreij: Always try to avoid `!!`, here's a suggestion of how to write this:

```kotlin
        if (paymentComponentState is CardComponentState) {
            val cardType = paymentComponentState.cardType?.txVariant
            paymentComponentJson.putOpt(BRAND_KEY, cardType)
        }
```

descorp: Fixed [57cc50a](https://github.com/Adyen/adyen-react-native/pull/210/commits/57cc50aac2ae93d627c6c32a048c3919db15d5e2)

RyanLinXiang (commented):
Thanks for opening this pull request - that is exactly what we need. But I tested on my end and it does not include the 'brand' key

descorp:
@RyanLinXiang are you testing on TEST or on LIVE ?

on TEST our "BinLookup" table is very basic and does not contain every card. Also make sure that network scheme is enabled on your "Customer Area".

You can try with this test card:

```
5454 5454 5454 5454
03/30     737
```

android/src/main/java/com/adyenreactnativesdk/component/dropin/AdyenCheckoutService.kt
```diff
@@ -30,12 +26,19 @@ open class AdyenCheckoutService : DropInService(), DropInModuleListener {
         paymentComponentState: PaymentComponentState<*>,
         paymentComponentJson: JSONObject
     ) {
+        if (paymentComponentState is CardComponentState &&
+            paymentComponentJson.getJSONObject(PAYMENT_DETAILS_KEY).isNull(BRAND_KEY)
+        ) {
+            val cardType = paymentComponentState.cardType?.txVariant
+            paymentComponentJson.getJSONObject(PAYMENT_DETAILS_KEY).putOpt(BRAND_KEY, cardType)
```
jreij, line 33: Did you make sure this actually ends up updating the `paymentComponentJson` param? Pretty sure you tested it but just a sanity check, I'm afraid whether `getJSONObject` returns a copy or the JSON object itself by reference.

descorp: Yep. If card recognised by BinLookup - the `brand` is populated.

Screenshot 2023-06-01 at 10 28 01

jreij (approved):

PR #247, opened 2023-07-12
Fixes #246 - crash when selecting Online Banking PL payment method

### LINKED ISSUE #246: [ANDROID] App got crashed when selecting Online Banking PL payment method
**Getting crash when selecting the payment method Online Banking PL**

**To Reproduce**

When the payment options were shown, selected the payment method Online Banking PL, app got crashed. Please refer the screenshot attached,

**Expected behavior**
Crash shouldn't happened when selecting the payment method.
**Screenshots**
image

**Smartphone (please complete the following information):**
 - Device: Android Emulator - Pixel 4(Android 11) API level 30. 
 - OS: Android

**Additional context**
 This crash happens only on android. Thanks.

moaeadw:
## Summary
The fields `NATIVE_COMPONENTS` and `UNSUPPORTED_PAYMENT_METHODS` contains uppercase letters such as `mealVoucher_FR_natixis, onlineBanking_PL`

but the field that is comparing it is lowered case 
`const type = typeName.toLowerCase();`

So the fix is just to use `typeName `which contains the original type

**Fixed issue**:  #246

### DIFF
--- a/src/AdyenNativeModules.tsx
+++ b/src/AdyenNativeModules.tsx
@@ -182,11 +182,11 @@ export function getNativeComponent(
     throw new Error(UNKNOWN_PAYMENT_METHOD_ERROR + typeName);
   }
 
-  if (UNSUPPORTED_PAYMENT_METHODS.includes(type)) {
+  if (UNSUPPORTED_PAYMENT_METHODS.includes(typeName)) {
     throw new Error(UNSUPPORTED_PAYMENT_METHOD_ERROR + typeName);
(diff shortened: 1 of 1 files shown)


ahsan-ali-lebara (approved):

descorp:
Thanks @moaead 💚

We would like to approach this problem from [another direction](https://github.com/Adyen/adyen-react-native/pull/250), but huge thanks for the tip on how to fix it :)

Issue #8, opened 2022-04-19
[NEED ASSIST] Customization options in Adyen3DS Screen


ahmetgsu:
In `adyen-3ds2-ios` and `adyen-3ds2-android` repositories, we can customize header text and color.
Can we do that in `adyen-react-native` repo? If possible, could you please guide me on how to do that?

![2022-04-19_10-21-47](https://user-images.githubusercontent.com/48296634/163958998-81daa865-55e5-4ad6-ad00-e1c01837a09e.png)

descorp:
Hey @ahmetgsu 

Thanks for reaching out!
Not yet, unfortunately.
Customisation and styling are in our backlog - right now we are gathering feedback to prioritise and decide what to work next.

arnaudschlupp:
Hi, regarding customization, on the Android side, adding the correct XML entries works quite fine on our side. Now regarding iOS, I wonder if we could have a similar solution by providing a class like this:

```
import Foundation
import IdensicMobileSDK

class IdensicMobileSDKCustomization: NSObject {

  @objc static func apply(_ sdk: SNSMobileSDK) {
    
    let primaryColor = UIColor(red: 229.0/255.0, green: 25.0/255.0, blue: 25.0/255.0, alpha: 1.0)
    let primaryDarkColor = UIColor(red: 0.77, green: 0.07, blue: 0.13, alpha: 1.00)
    let primaryContrastColor = UIColor.white
    let grey50Color = UIColor(white: 0.5, alpha: 1.0)
...
```

We use this solution for another package we are using in our app. It does the trick. Of course, a style object where you can map everything would be even handier.

descorp:
Hey @arnaudschlupp 

Customisation is indeed a pressing issue, since iOS does not have static overridable resources like Android does. 

> providing a class like this

Since we have to handle customisations per platform, it could be a feasible solution.
Thanks for a very fresh idea 💚

iyanushow:
Hi @descorp @arnaudschlupp, I'm also looking to style the checkout component, specifically the accent color and the pay button background color so it aligns with our app design . Is there a way we could do this? 

I came across the AdyenAppearance swift class, 
```
import Foundation
import Adyen
import adyen_react_native

/// Please make sure that the name of the class exactly matches.
/// SDK will use reflection to find the class with this exact name.
class AdyenAppearance: AdyenAppearanceProvider {
  static func createStyle() -> Adyen.DropInComponent.Style {
[...]

PR #253, opened 2023-07-18
fix: Use CFBundleName rather than CFBundleIdentifier for finding the …

slhodson969:
…AdyenAppearance class

## Summary
Use CFBundleName rather than CFBundleIdentifier for finding the AdyenAppearance class

## Tested scenarios
N/A

**Fixed issue**:  

https://github.com/Adyen/adyen-react-native/issues/8#issuecomment-1620102660


descorp:
Great thanks for your quick PR @slhodson969  💚

Unfortunately, I am going to close it since we are going to address this issue with slightly different approach #255 
Hope this will not discourage from future contributions!

PR #260, opened 2023-08-06
Fix android crash on invalid actiondata

stefkampen:
## Summary
When the handle function was invoked from an invalid action response the android app would crash.
In my case this would result in a fatal crash with the error description:
sdkData not found for type paymentMethodType - paypal

Changing the catch handler to catch all exceptions fixes this crash.


descorp (approved):

descorp:
Thank you for contribution @stefkampen  💚

PR #337, opened 2024-01-12
Avoid Render Key Conflict error message in example app

StefanWallin:
## Summary
When adding f.x. two cards from same issuer(mastercard, visa, amex et.c.), this view key would end up being the same, and React keys have to be unique. This caused the second card ending in 6974 being hid from the list in the example app.

Screenshot 2024-01-12 at 09 34 03

## Tested scenarios
Tested adding multiple cards:

| Before | After |
| --- | --- |
| Screenshot 2024-01-12 at 09 35 26 | Screenshot 2024-01-12 at 09 35 46 |


example/src/Views/Checkout/PaymentMethodsView.js
```diff
@@ -69,7 +69,7 @@ const PaymentMethods = () => {
         {regularPaymentMethods.map((p) => {
           const iconName = p.type === 'scheme' ? 'card' : p.type;
           return (
-            <View key={`${p.type + p.name}`}>
+            <View key={`regular-payment-method-${p.id}`}>
```
descorp: Regular payment methods don't have `id`.

```suggestion
            <View key={`${p.type + p.name}`}>
```

example/src/Views/Checkout/PaymentMethodsView.js
```diff
@@ -50,7 +50,7 @@ const PaymentMethods = () => {
             {storedPaymentMethods.map((p) => {
               const iconName = p.type === 'scheme' ? 'card' : p.type;
               return (
-                <View key={`${p.type + p.name}`}>
+                <View key={`stored-payment-method-${p.id}`}>
```
descorp:
```suggestion
                <View key={`${p.id}`}>
```

descorp (approved):

descorp:
Thanks @StefanWallin 💚

PR #338, opened 2024-01-15
Standalone action handling

descorp:
# Open PR

## Changes

* add standalone action handling via `AdyenAction` module for redirect and 3DS
* updater CSE example to work with 3DS2 action handling
* format JS

### DIFF
--- a/android/src/main/java/com/adyenreactnativesdk/component/base/BaseModule.kt
+++ b/android/src/main/java/com/adyenreactnativesdk/component/base/BaseModule.kt
@@ -13,6 +13,7 @@ import android.os.Build
 import androidx.appcompat.app.AppCompatActivity
 import androidx.lifecycle.lifecycleScope
 import com.adyen.checkout.action.core.internal.ActionHandlingComponent
+import com.adyen.checkout.adyen3ds2.Cancelled3DS2Exception
 import com.adyen.checkout.components.core.PaymentComponentData
 import com.adyen.checkout.components.core.PaymentComponentState
 import com.adyen.checkout.components.core.PaymentMethod
@@ -153,7 +154,10 @@ abstract class BaseModule(context: ReactApplicationContext?) : ReactContextBaseJ
     }
 
     override fun onException(exception: CheckoutException) {
-        if (exception is CancellationException || exception.message == "Payment canceled.") {
+        if (exception is CancellationException ||
+            exception is Cancelled3DS2Exception ||
+            exception.message == "Payment canceled."
+        ) {
             sendErrorEvent(ModuleException.Canceled())
         } else {
             sendErrorEvent(exception)
--- a/android/src/main/java/com/adyenreactnativesdk/cse/ActionFragment.kt
+++ b/android/src/main/java/com/adyenreactnativesdk/cse/ActionFragment.kt
@@ -1,0 +1,99 @@
+package com.adyenreactnativesdk.cse;
+
+import android.os.Bundle
+import android.util.Log
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import androidx.fragment.app.FragmentManager
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.lifecycleScope
+import androidx.lifecycle.repeatOnLifecycle
+import com.adyen.checkout.action.core.GenericActionComponent
+import com.adyen.checkout.action.core.GenericActionConfiguration
+import com.adyen.checkout.components.core.ActionComponentCallback
+import com.adyen.checkout.components.core.ActionComponentData
+import com.adyen.checkout.components.core.action.Action
+import com.adyen.checkout.ui.core.AdyenComponentView
+import com.adyenreactnativesdk.AdyenCheckout
+import com.adyenreactnativesdk.R
+import com.google.android.material.bottomsheet.BottomSheetDialogFragment
+import kotlinx.coroutines.launch
+
+class ActionFragment(
+    private val configuration: GenericActionConfiguration,
+    private val callback: ActionComponentCallback,
+    private val action: Action,
+) :
+    BottomSheetDialogFragment() {
+
+    private var actionHandled: Boolean = false
+    var component: GenericActionComponent? = null
+
+    override fun onCreateView(
+        inflater: LayoutInflater,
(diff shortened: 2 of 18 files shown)


goergisn (dismissed):

example/src/Utilities/Helpers.js
```diff
@@ -0,0 +1,12 @@
+import { ResultCode } from '@adyen/react-native';
+
+export const isSuccess = (
```
longyulongyu: It's optional, it can be simplified as:
```javascript
export const isSuccess = ({ resultCode }) =>
    [ResultCode.authorised, ResultCode.received, ResultCode.pending, ResultCode.presentToShopper].includes(resultCode);
```

example/src/Views/CseView.js
```diff
@@ -43,21 +46,28 @@ const CseView = ({ navigation }) => {
           encryptedExpiryMonth: encryptedCard.expiryMonth,
           encryptedExpiryYear: encryptedCard.expiryYear,
           encryptedSecurityCode: encryptedCard.cvv,
+          threeDS2SdkVersion: threeDS2SdkVersion,
         },
       };
 
-      const result = await ApiClient.payments(data, configuration);
+      var result = await ApiClient.payments(data, configuration, "myapp://payments");
```
longyulongyu: I think we can use `let`.

descorp: It is overwritten later on (LINE 57).

`result` here is `\payment` API response. If payment returns no action - it proceeds, if there is `action - it enters **IF** and rewrites `result` with `\paymentDetails` API response.

example/src/Utilities/Helpers.js
descorp: Fixed [1ba5c8d](https://github.com/Adyen/adyen-react-native/pull/338/commits/1ba5c8db1f7f9340e531502da1414fe7a1124558)

longyulongyu (dismissed):

android/src/main/java/com/adyenreactnativesdk/component/base/BaseModule.kt
```diff
@@ -153,7 +154,10 @@ abstract class BaseModule(context: ReactApplicationContext?) : ReactContextBaseJ
     }
 
     override fun onException(exception: CheckoutException) {
-        if (exception is CancellationException || exception.message == "Payment canceled.") {
+        if (exception is CancellationException ||
+            exception is Cancelled3DS2Exception ||
+            exception.message == "Payment canceled."
```
Robert-SD, line 159: Is it possible that the string might change at runtime? If so, is the normal exception sufficient?

android/src/main/java/com/adyenreactnativesdk/cse/ActionFragment.kt
```diff
@@ -0,0 +1,99 @@
...
+        savedInstanceState: Bundle?
+    ): View {
+        dialog?.setCanceledOnTouchOutside(false)
+        return inflater.inflate(R.layout.fragment_instant, container)
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        viewLifecycleOwner.lifecycleScope.launch {
+            repeatOnLifecycle(Lifecycle.State.STARTED) {
```
Robert-SD: Let's discuss if we can use the onStart() callback of the fragment instead of the onViewCreated. Maybe we can just call setupComponent() in the onStart() callback.

android/src/main/java/com/adyenreactnativesdk/cse/ActionFragment.kt
```diff
@@ -0,0 +1,99 @@
...
+                launch {
+                    setupComponent(action)
+                }
+            }
+        }
+    }
+
+    private fun setupComponent(action: Action) {
+        ActionComponentData
```
Robert-SD: Do we need to delete this line?

android/src/main/java/com/adyenreactnativesdk/cse/ActionModule.kt
```diff
@@ -0,0 +1,108 @@
...
+
+    override fun onAdditionalDetails(actionComponentData: ActionComponentData) {
+        val json = ActionComponentData.SERIALIZER.serialize(actionComponentData)
+        promise?.resolve(ReactNativeJson.convertJsonToMap(json))
+    }
+
+    override fun onError(componentError: ComponentError) {
+        if (componentError.exception is CancellationException ||
+            componentError.exception is Cancelled3DS2Exception ||
+            componentError.exception.message == "Payment canceled."
```
Robert-SD: This is the same string. Shall we extract it into a const for reuse?

descorp: Actually not needed here
[8ab85a2](https://github.com/Adyen/adyen-react-native/pull/338/commits/8ab85a203df1504c6054d14871a372b88a9b50e8)

android/src/main/java/com/adyenreactnativesdk/cse/ActionFragment.kt
descorp: Fixed [eb8d84c](https://github.com/Adyen/adyen-react-native/pull/338/commits/eb8d84c3cf6a325641d6cd41576f76bb762b1c87)

android/src/main/java/com/adyenreactnativesdk/cse/ActionFragment.kt
descorp: Fixed [eb8d84c](https://github.com/Adyen/adyen-react-native/pull/338/commits/eb8d84c3cf6a325641d6cd41576f76bb762b1c87)

android/src/main/java/com/adyenreactnativesdk/component/base/BaseModule.kt
descorp: This is how [GooglePayComponent](https://github.com/Adyen/adyen-android/blob/develop/googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/DefaultGooglePayDelegate.kt#L156) throws it..
This can actually be `CancellationException`

Robert-SD (approved):

PR #341, opened 2024-01-18
Refactor AdyenNativeModules.js

descorp:
# Open PR

## Changes

* Move `AdyenNativeComponentWrapper` to a standalone file
* Move `getNativeComponent` to a standalone file
* Add `ActionModuleWrapper` to handle `threeDS2SdkVersion ` constant

### DIFF
--- a/src/getNativeComponent.tsx
+++ b/src/getNativeComponent.tsx
@@ -26,7 +26,7 @@ export function getNativeComponent(
   paymentMethods: PaymentMethodsResponse,
 ): {
   nativeComponent: AdyenActionComponent & NativeModule;
-  paymentMethod: PaymentMethod | undefined;
+  paymentMethod?: PaymentMethod;
 } {
   switch (typeName) {
     case 'dropin':
@@ -37,7 +37,6 @@ export function getNativeComponent(
         nativeComponent: new AdyenNativeComponentWrapper({
           nativeModule: AdyenDropIn,
         }),
-        paymentMethod: undefined,
       };
     case 'applepay':
       return {
@@ -45,7 +44,6 @@ export function getNativeComponent(
           nativeModule: AdyenApplePay,
           canHandleAction: false,
         }),
-        paymentMethod: undefined,
       };
     case 'paywithgoogle':
     case 'googlepay':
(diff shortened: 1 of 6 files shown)


src/getNativeComponent.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Get native component capable of handling provided payment method type.
+ */
+export function getNativeComponent(
+  typeName: string,
+  paymentMethods: PaymentMethodsResponse,
+): {
+  nativeComponent: AdyenActionComponent & NativeModule;
+  paymentMethod: PaymentMethod | undefined;
```
longyulongyu: I think we can define `paymentMethod: PaymentMethod | undefined;` as optional? `paymentMethod?: PaymentMethod;`

src/getNativeComponent.tsx
```diff
@@ -0,0 +1,86 @@
...
+  switch (typeName) {
+    case 'dropin':
+    case 'dropIn':
+    case 'drop-in':
+    case 'adyendropin':
+      return {
+        nativeComponent: new AdyenNativeComponentWrapper({
+          nativeModule: AdyenDropIn,
+        }),
+        paymentMethod: undefined,
```
longyulongyu: If we define `paymentMethod?: PaymentMethod;` as optional, we don't need to return `paymentMethod: undefined`. Same for other cases.
So it would be 
```
return {
        nativeComponent: new AdyenNativeComponentWrapper({
          nativeModule: AdyenDropIn,
        })
      };
```

descorp: Fixed [0cb5a11](https://github.com/Adyen/adyen-react-native/pull/341/commits/0cb5a115d0c3a9f599933c4ab96045713f5ceb49)

src/getNativeComponent.tsx
descorp: Fixed [0cb5a11](https://github.com/Adyen/adyen-react-native/pull/341/commits/0cb5a115d0c3a9f599933c4ab96045713f5ceb49)

longyulongyu (approved):

PR #425, opened 2024-04-17
Add Long support

descorp:
# Open PR

## Changes

* allow Long smaller then `Int.MAX_VALUE` treated as numbers

### DIFF
--- a/android/src/main/java/com/adyenreactnativesdk/util/ReactNativeJson.kt
+++ b/android/src/main/java/com/adyenreactnativesdk/util/ReactNativeJson.kt
@@ -26,22 +26,15 @@ object ReactNativeJson {
         while (iterator.hasNext()) {
             val key = iterator.next()
             val value = jsonObject[key]
-            if (value is JSONObject) {
-                map.putMap(key, convertJsonToMap(value))
-            } else if (value is JSONArray) {
-                map.putArray(key, convertJsonToArray(value))
-            } else if (value is Boolean) {
-                map.putBoolean(key, value)
-            } else if (value is Long && value <= Int.MAX_VALUE) {
-                map.putInt(key, value.toInt())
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/adyenreactnativesdk/util/ReactNativeJson.kt
```diff
@@ -32,6 +32,8 @@ object ReactNativeJson {
                 map.putArray(key, convertJsonToArray(value))
             } else if (value is Boolean) {
                 map.putBoolean(key, value)
+            } else if (value is Long && (value as Long) < Int.MAX_VALUE) {
```
jreij: I think `(value as Long)` is redundant because of `value is Long` before it, can you double check?

descorp: Indeed :)
Fixed [3732d96](https://github.com/Adyen/adyen-react-native/pull/425/commits/3732d964adb1012a688d041bbc6e8cb6f463d80c)

android/src/main/java/com/adyenreactnativesdk/util/ReactNativeJson.kt
jreij, line 1: Minor suggestion, you can replace your `if...else if...` with a `when` to improve readability. Just put the cursor on the first `if`, press `command + enter` and `replace if with when`.

Screenshot 2024-04-17 at 14 33 05

descorp: Fixed [6fefd45](https://github.com/Adyen/adyen-react-native/pull/425/commits/6fefd4572eb5ba8e8cbd76d1480f26c2a448cf01)

jreij (approved):

PR #455, opened 2024-05-30
Improve expo-router compatibility

descorp:
# Open PR

## Changes

* Now expo plugin generates `AppDelegate` code in a way that do not cause conflicts with deep linking.


kpoelhekke:
Maybe modify the Manual integration readme too. I guess this is also an issue for non-expo implementations?

Robert-SD (dismissed):

Robert-SD (approved):

PR #542, opened 2024-10-23
Partial payments

descorp:
# Open PR

## Changes

* add Partial payment support for Drop-In
* rename "didDisableStoredPaymentMethod" to "didDisableStoredPaymentMethodCallback" to align with the rest
* refactor `example/src/AppContext.js`
* fix example app bug when amount was not saving in Settings

### DIFF
--- a/android/src/main/java/com/adyenreactnativesdk/component/dropin/AdvancedCheckoutService.kt
+++ b/android/src/main/java/com/adyenreactnativesdk/component/dropin/AdvancedCheckoutService.kt
@@ -9,6 +9,7 @@ package com.adyenreactnativesdk.component.dropin
 import android.util.Log
 import com.adyen.checkout.components.core.ActionComponentData
 import com.adyen.checkout.components.core.LookupAddress
+import com.adyen.checkout.components.core.Order
 import com.adyen.checkout.components.core.PaymentComponentState
 import com.adyen.checkout.dropin.DropInService
 import com.adyenreactnativesdk.AdyenCheckout
@@ -32,8 +33,7 @@ open class AdvancedCheckoutService : DropInService() {
 
     override fun onAdditionalDetails(actionComponentData: ActionComponentData) {
         val listener = CheckoutProxy.shared.componentListener
-        val actionComponentJson = ActionComponentData.SERIALIZER.serialize(actionComponentData)
-        listener?.onAdditionalData(actionComponentJson)
+        listener?.onAdditionalDetails(actionComponentData)
             ?: Log.e(
                 TAG,
                 "Invalid state: DropInServiceListener is missing"
@@ -46,6 +46,21 @@ open class AdvancedCheckoutService : DropInService() {
 
     override fun onAddressLookupCompletion(lookupAddress: LookupAddress): Boolean {
         return CheckoutProxy.shared.addressLookupCallback?.onLookupCompletion(lookupAddress) ?: false
+    }
+
+    override fun onBalanceCheck(paymentComponentState: PaymentComponentState<*>) {
+        val listener = CheckoutProxy.shared.componentListener
+        listener?.onBalanceCheck(paymentComponentState)
+    }
+
+    override fun onOrderRequest() {
+        val listener = CheckoutProxy.shared.componentListener
+        listener?.onOrderRequest()
+    }
+
+    override fun onOrderCancel(order: Order, shouldUpdatePaymentMethods: Boolean) {
+        val listener = CheckoutProxy.shared.componentListener
+        listener?.onOrderCancel(order, false)
     }
 
     companion object {
--- a/android/src/main/java/com/adyenreactnativesdk/component/dropin/AdvancedCheckoutService.kt
+++ b/android/src/main/java/com/adyenreactnativesdk/component/dropin/AdvancedCheckoutService.kt
@@ -60,7 +60,7 @@ open class AdvancedCheckoutService : DropInService() {
 
     override fun onOrderCancel(order: Order, shouldUpdatePaymentMethods: Boolean) {
         val listener = CheckoutProxy.shared.componentListener
-        listener?.onOrderCancel(order, false)
+        listener?.onOrderCancel(order, shouldUpdatePaymentMethods)
     }
 
     companion object {
--- a/android/src/main/java/com/adyenreactnativesdk/component/dropin/DropInModule.kt
+++ b/android/src/main/java/com/adyenreactnativesdk/component/dropin/DropInModule.kt
@@ -213,7 +213,7 @@ class DropInModule(context: ReactApplicationContext?) : BaseModule(context), Rea
             listener.sendBalanceResult(BalanceDropInServiceResult.Balance(balanceResult))
         } else {
             val message = error?.getString(AdyenConstants.PARAMETER_MESSAGE)
-            listener.sendBalanceResult(BalanceDropInServiceResult.Error(null, message, false)  )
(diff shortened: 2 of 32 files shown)


ios/Components/DropInModule.swift
```diff
@@ -47,7 +47,7 @@ internal final class DropInModule: BaseModule {
             }
 
             do {
-                let address: LookupAddressModel = try address.toJson()
+                let address: LookupAddressModel = try address.decode()
```
goergisn: Nitpic: Variable shadowing
```suggestion
                let addressModel: LookupAddressModel = try address.decode()
```

goergisn (dismissed):

example/src/Utilities/APIClient.js
```diff
@@ -36,11 +36,12 @@ class ApiClient {
     return ApiClient.makeRequest(ENVIRONMENT.url + 'sessions', body);
   };
 
-  static paymentMethods = configuration => {
+  static paymentMethods = (configuration, order) => {
     const body = {
       ...parseConfig(configuration),
       ...parseAmount(configuration),
       ...serverConfiguration,
+      order: order ? parseOrder(order) : undefined,
```
longyulongyu: I think we can omit the order key if there is no `order`. So that we don't include `order: undefined` in the body.

```js
...(order && { order: parseOrder(order) }),
```

descorp: Fixed on [4513c5b](https://github.com/Adyen/adyen-react-native/pull/542/commits/4513c5b173e7e9d053e941bfc33777b423d7404a)

ios/Components/DropInModule.swift
descorp: Fixed on [4513c5b](https://github.com/Adyen/adyen-react-native/pull/542/commits/4513c5b173e7e9d053e941bfc33777b423d7404a)

longyulongyu (dismissed):

OscarSpruit (dismissed):

erenbesel (dismissed):

android/src/main/java/com/adyenreactnativesdk/component/dropin/AdvancedCheckoutService.kt
```diff
@@ -48,6 +48,21 @@ open class AdvancedCheckoutService : DropInService() {
...
+    }
+
+    override fun onOrderRequest() {
+        val listener = CheckoutProxy.shared.componentListener
+        listener?.onOrderRequest()
+    }
+
+    override fun onOrderCancel(order: Order, shouldUpdatePaymentMethods: Boolean) {
+        val listener = CheckoutProxy.shared.componentListener
+        listener?.onOrderCancel(order, false)
```
Robert-SD:
```suggestion
        listener?.onOrderCancel(order, shouldUpdatePaymentMethods)
```

android/src/main/java/com/adyenreactnativesdk/component/dropin/DropInModule.kt
```diff
@@ -195,6 +200,57 @@ class DropInModule(context: ReactApplicationContext?) : BaseModule(context), Rea
...
+            sendErrorEvent(ModuleException.NoModuleListener(integration))
+            return
+        }
+        if (success) {
+            val jsonObject = ReactNativeJson.convertMapToJson(balance)
+            val balanceResult = BalanceResult.SERIALIZER.deserialize(jsonObject)
+            listener.sendBalanceResult(BalanceDropInServiceResult.Balance(balanceResult))
+        } else {
+            val message = error?.getString(AdyenConstants.PARAMETER_MESSAGE)
+            listener.sendBalanceResult(BalanceDropInServiceResult.Error(null, message, true)  )
```
ozgur00, line 216: Did you leave first parameter (which is the error dialog that will be shown) in `BalanceDropInServiceResult.Error()` as null on purpose? I assume we're not showing any pop ups when this happens in the flow right now. On android we pass an `ErrorDialog` here when something goes wrong to show a pop up. You can take a look at `handleBalanceResponse` function in `ExampleAdvancedDropInService` or `requestBalanceCall` in `SessionDropInService` to see how we do it. The same goes for other kinds of `DropInServiceResult.Error` instances such as order creation etc.

android/src/main/java/com/adyenreactnativesdk/component/dropin/DropInModule.kt
```diff
@@ -195,6 +200,57 @@ class DropInModule(context: ReactApplicationContext?) : BaseModule(context), Rea
...
+            sendErrorEvent(ModuleException.NoModuleListener(integration))
+            return
+        }
+        if (success) {
+            val jsonObject = ReactNativeJson.convertMapToJson(order)
+            val orderResponse = OrderResponse.SERIALIZER.deserialize(jsonObject)
+            listener.sendOrderResult(OrderDropInServiceResult.OrderCreated(orderResponse))
+        } else {
+            val message = error?.getString(AdyenConstants.PARAMETER_MESSAGE)
+            listener.sendOrderResult(OrderDropInServiceResult.Error(null, message, true)  )
```
ozgur00, line 233: Also the last parameter here is `dismissDropIn` which is by default `false`. Is it intentionally set to `true` here? That will dismiss drop-in automatically, either immediately or if a pop-up is shown after the dismissal of that. If it's not intentional I suggest not passing it so Android SDK would decide what to do which would be not dismissing drop-in.

android/src/main/java/com/adyenreactnativesdk/component/dropin/DropInModule.kt
descorp: Correct. Intent here to dismiss the DropIn and communicate error back to merchant's level and so they can handle UI state (present alert etc)
Another reason - there is no popup on iOS, so dismissing DropIn makes behaviour more align.

android/src/main/java/com/adyenreactnativesdk/component/dropin/DropInModule.kt
descorp: Related to https://github.com/Adyen/adyen-react-native/pull/542#discussion_r1815057166

ozgur00 (dismissed):

longyulongyu (approved):

android/src/main/java/com/adyenreactnativesdk/component/dropin/AdvancedCheckoutService.kt
descorp: Resolved [1b1f9eb](https://github.com/Adyen/adyen-react-native/pull/542/commits/1b1f9eb3e9cee1b01a4a1d9778424fa292897ad6)

example/src/Views/Checkout/PartialPaymentCheckout.js
```diff
@@ -164,18 +162,26 @@ const PartialPaymentCheckout = ({ navigation }) => {
...
+  const cancelOrder = useCallback(
+    async (
+      /** @type {import('@adyen/react-native').Order} */ order, 
+      /** @type {Boolean} */ shouldUpdatePaymentMethods,
+      /** @type {import('@adyen/react-native').DropInModule} */ dropInComponent) => {
       try {
         await ApiClient.cancelOrder(order, configuration);
+        if (shouldUpdatePaymentMethods) {
+          let paymentMethods = await ApiClient.paymentMethods(configuration, order);
+          dropInComponent.providePaymentMethods(paymentMethods, undefined);
```
Robert-SD, line 181: Do you need to pass the order here or should it be undefined?

Robert-SD (approved):

PR #546, opened 2024-11-05
Hide pin

descorp:
# Open PR

## Changes

* add `pinRequired` property to partial payment configuration
* add iOS tests for root configuration and analytics
* defining analytics as ON unless set otherwise

### DIFF
--- a/ios/Configuration/PartialPaymentParser.swift
+++ b/ios/Configuration/PartialPaymentParser.swift
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2022 Adyen N.V.
+// Copyright (c) 2024 Adyen N.V.
 //
 // This file is open source and available under the MIT license. See the LICENSE file for more info.
 //
--- a/ios/Configuration/PartialPaymentParser.swift
+++ b/ios/Configuration/PartialPaymentParser.swift
@@ -1,0 +1,36 @@
+//
+// Copyright (c) 2022 Adyen N.V.
+//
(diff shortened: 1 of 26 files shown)


OscarSpruit (dismissed):

ios/Configuration/PartialPaymentParser.swift
```diff
@@ -0,0 +1,36 @@
...
+
+public struct PartialPaymentParser {
+
+    private var dict: [String: Any]
+
+    init(configuration: [String: Any]) {
+        self.init(configuration: configuration as NSDictionary)
+    }
+
+    public init(configuration: NSDictionary) {
```
atmamont, line 13: Any specific reason to use `NSDictionary`?

descorp: ReactNative [operates with NSDictionary](https://reactnative.dev/docs/legacy/native-modules-ios#argument-types) since it is bridging JS via Obj-c

atmamont: Got it!

I see there are 25 occurrences of casting to `NSDictionary`, and I see there is an internal  `init(configuration: [String: Any])`. Would it be possible to remove excessive casting from unit tests part?

descorp: Fixed [5ab7744](https://github.com/Adyen/adyen-react-native/pull/546/commits/5ab774430ba8312d7df2c3cd97094195b042a58e)

atmamont (approved):

PR #564, opened 2024-12-27
fix: override fun onNewIntent not nullable param

lucashe1997:
## Summary
While trying to implement this package, I encountered a problem when using Expo 52 with a development build, specifically with the Kotlin language during the Android build:
```shell
Android gradle plugin: 8.6.0
Gradle: 8.10.2

> Task :app:compileDebugKotlin FAILED
e: file:///.../android/app/src/main/java/.../MainActivity.kt:69:3 'onNewIntent' overrides nothing
e: file:///.../android/app/src/main/java/.../MainActivity.kt:70:23 Type mismatch: inferred type is Intent? but Intent was expected

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:compileDebugKotlin'.
> A failure occurred while executing org.jetbrains.kotlin.compilerRunner.GradleCompilerRunnerWithWorkers$GradleKotlinCompilerWorkAction
   > Compilation error. See log for more details
``` 

After investigating the issue, I found that the problem originates from the following piece of code:
```kotlin
  override fun onNewIntent(intent: Intent?) {
    super.onNewIntent(intent)
    intent?.let { AdyenCheckout.handleIntent(it) }
  }
``` 

The Android documentation for [onNewIntent](https://developer.android.com/reference/android/app/Activity#onNewIntent(android.content.Intent)) specifies that the `intent` parameter is non-nullable. As a result, the null check (`intent?.let`) is unnecessary. This behavior is consistent with the Java implementation provided in the same plugin:

```java
  @Override
  public void onNewIntent(Intent intent) {
    super.onNewIntent(intent);
    AdyenCheckout.handleIntent(intent);
  }
```

### DIFF
--- a/plugin/src/setKotlinMainActivity.ts
+++ b/plugin/src/setKotlinMainActivity.ts
@@ -11,18 +11,18 @@ export function setKotlinMainActivity(contents: string): string {
   );
 
   // on NewIntent
-  if (contents.includes('override fun onNewIntent(intent: Intent?) {')) {
+  if (contents.includes('override fun onNewIntent(intent: Intent) {')) {
     contents = contents.replace(
       'super.onNewIntent(intent)\n',
-      'super.onNewIntent(intent)\n    intent?.let { AdyenCheckout.handleIntent(it) }\n'
+      'super.onNewIntent(intent)\n    AdyenCheckout.handleIntent(intent)\n'
     );
   } else {
     contents = contents.replace(
       /}\n$/,
       '\n' +
-        '  override fun onNewIntent(intent: Intent?) {\n' +
+        '  override fun onNewIntent(intent: Intent) {\n' +
         '    super.onNewIntent(intent)\n' +
-        '    intent?.let { AdyenCheckout.handleIntent(it) }\n' +
+        '    AdyenCheckout.handleIntent(intent)\n' +
         '  }\n' +
         '}\n'
     );


Zorig:
faced the same issue @jreij

descorp:
Hey @lucashe1997 

Thank you for making PR! 💚

This breaking change introduced by Google on Android [API level 35](https://developer.android.com/reference/android/app/Activity#onNewIntent(android.content.Intent,%20android.app.ComponentCaller)). 

Before we merge this pull request, we ideally need to ensure support for older API levels as well. I am currently exploring ways to achieve this.

Alternatively, we are considering changing the implementation to use a `Consumer<Intent>` instead of overriding `onNewIntent`, but this will require a major version release.

robyhz:
any update ?

vanGalilea (approved):

vanGalilea:
@descorp can we expedite a fix release for this. please?

descorp:
Hey @vanGalilea 

Merging this pull request will cause integration issues for anyone who is not using API level 35 yet. Unfortunately, we will need to ask you to implement a manual fix until we can find a comprehensive solution or until API level 35 becomes widely adopted.

Thanks for checking!

lucashe1997:
Hi @vanGalilea as @descorp mentioned this PR solves the issue for the API level 35 and later, but it'll be a problem with older APIs.

If you are using expo, you can apply get the patch file https://patch-diff.githubusercontent.com/raw/Adyen/adyen-react-native/pull/564.patch and apply it using patch-package for example. It not you can change the code manually.

@descorp if this PR is causing too much confusion we can close it, and I can create just a issue, with the patch. Just let me know!

descorp:
Thanks @lucashe1997  💚

PR is perfect! :D
We will address this issue eventually. It is not a priority right now since it is not a blocker and a workaround is available.

vanGalilea:
Something didn't work with the suggested patch. I suspect this early [exit](https://github.com/Adyen/adyen-react-native/blob/cf04d8b59e877792aaad37a75ed51547ec39c93e/plugin/src/withAdyenAndroid.ts#L13) in `withAdyenAndroid.ts`

I solved this by applying the following custom plugin, see [gist](https://gist.github.com/vanGalilea/59c149a6c11c78082a4f4aaaae4fc993).

andrico1234:
> Something didn't work with the suggested patch. I suspect this early [exit](https://github.com/Adyen/adyen-react-native/blob/cf04d8b59e877792aaad37a75ed51547ec39c93e/plugin/src/withAdyenAndroid.ts#L13) in `withAdyenAndroid.ts`
> 
> I solved this by applying the following custom plugin, see [gist](https://gist.github.com/vanGalilea/59c149a6c11c78082a4f4aaaae4fc993).

Sorry if off topic, but I just wanted to thank you for this solution, it worked a treat!

I also tested this with Android API 34 and it's working as expected

descorp:
I am going to close this PR in favour of #576

PR #576, opened 2025-03-04
Adjust Expo prebuild for AndroidAPI level 50 and 52

descorp:
# Open PR

## Changes

- correct `onNewIntent` override used for Expo version 50 and above
- deprecated `onActivityResult` no longer used for Expo 52 and above

### DIFF
--- a/plugin/src/withAdyenAndroid.ts
+++ b/plugin/src/withAdyenAndroid.ts
@@ -16,7 +16,8 @@ export const withAdyenAndroid: ConfigPlugin = (config) => {
     if (mainActivity.language === 'java') {
       mainActivity.contents = setJavaMainActivity(mainActivity.contents);
     } else {
-      mainActivity.contents = setKotlinMainActivity(mainActivity.contents);
+      const sdkVersion = parseInt(config.sdkVersion.split('.')[0]);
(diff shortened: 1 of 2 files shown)


plugin/src/withAdyenAndroid.ts
```diff
@@ -16,7 +16,12 @@ export const withAdyenAndroid: ConfigPlugin = (config) => {
     if (mainActivity.language === 'java') {
       mainActivity.contents = setJavaMainActivity(mainActivity.contents);
     } else {
-      mainActivity.contents = setKotlinMainActivity(mainActivity.contents);
+      const sdkVersion = config.sdkVersion ?? '49.0.0';
```
Robert-SD, line 19: Could a provided sdkVersion through a config contain invalid characters like a dash or colon? If so, is the parsing safe or could it throw an exception?

Robert-SD (approved):

plugin/src/withAdyenAndroid.ts
descorp: Version is semver and can look like `53.0.0-canary-20250304-f08e984` or `52.0.0-preview.13`, but parsing should still work.

Everything is possible, but value comes from Expo, so I hope it is correct 😅

PR #598, opened 2025-04-25
Handle iDEAL as Instant payment

descorp:
# Open PR

## Changes

* Handle iDEAL flow with `IdealComponent` instead of DropIn

### DIFF
--- a/android/src/main/java/com/adyenreactnativesdk/component/instant/IdealFragment.kt
+++ b/android/src/main/java/com/adyenreactnativesdk/component/instant/IdealFragment.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023 Adyen N.V.
+ * Copyright (c) 2025 Adyen N.V.
  *
  * This file is open source and available under the MIT license. See the LICENSE file for more info.
  *
@@ -13,6 +13,8 @@ import androidx.fragment.app.FragmentManager
 import com.adyen.checkout.components.core.CheckoutConfiguration
 import com.adyen.checkout.components.core.PaymentMethod
 import com.adyen.checkout.components.core.action.Action
+import com.adyen.checkout.ideal.IdealComponent
+import com.adyen.checkout.ideal.IdealComponentState
 import com.adyen.checkout.instant.InstantComponentState
 import com.adyen.checkout.instant.InstantPaymentComponent
 import com.adyen.checkout.sessions.core.CheckoutSession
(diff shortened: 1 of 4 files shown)


android/src/main/java/com/adyenreactnativesdk/component/instant/IdealFragment.kt
```diff
@@ -0,0 +1,85 @@
...
+import com.adyen.checkout.sessions.core.CheckoutSession
+import com.adyen.checkout.ui.core.AdyenComponentView
+import com.adyenreactnativesdk.AdyenCheckout
+import com.adyenreactnativesdk.R
+import com.adyenreactnativesdk.component.base.BaseComponentFragment
+import com.adyenreactnativesdk.component.base.ComponentData
+import com.adyenreactnativesdk.component.base.ModuleException
+
+class IdealFragment(
+    private val configuration: CheckoutConfiguration,
```
Robert-SD, line 29: Could it happen that Android recreates the IdealFragment due to lifecycle changes and then the constructor fails because of the missing parameters? I think we used only the companion object.

android/src/main/java/com/adyenreactnativesdk/component/instant/InstantModule.kt
```diff
@@ -45,7 +45,12 @@ class InstantModule(context: ReactApplicationContext?) : BaseModule(context),
         }
 
         CheckoutProxy.shared.componentListener = this
-        InstantFragment.show(
+        when (paymentMethod.type) {
+            "ideal" -> fragment = IdealFragment
```
Robert-SD: Shall we use PaymentMethodTypes.IDEAL instead of the hardcoded string?

android/src/main/java/com/adyenreactnativesdk/component/instant/IdealFragment.kt
OscarSpruit, line 29: Fragment and Activities should not have constructor params. When the system decides to rebuild the Fragment it will use the constructor without params at all times, so any constructor params you need will not be set.

To workaround this you could do something like this in the `show` method below:
```Kotlin
override fun show(
        fragmentManager: FragmentManager,
        configuration: CheckoutConfiguration,
        paymentMethod: PaymentMethod,
        session: CheckoutSession?
) {
    IdealFragment().apply {
            arguments = bundleOf(key to value, key to value....)
        }
        .show(fragmentManager, TAG)
}
```
and in the fragment you retrieve them with:
```Kotlin
private val something: String = arguments?.get("key")
```

descorp: This is part of a bigger refactoring. Will address in another PR.

descorp: This is part of a bigger refactoring. Will address in another PR.

android/src/main/java/com/adyenreactnativesdk/component/instant/InstantModule.kt
descorp: This is good one!

descorp: FIxed [cd26f60](https://github.com/Adyen/adyen-react-native/pull/598/commits/cd26f60a6e33e916fa81706675651d8c711c8cb9)

Robert-SD (approved):

Issue #606, opened 2025-05-12
Library does not compile for android on Expo 53 with CNG/Prebuild


StefanWallin:
**Describe the bug**
Using Expo 53 and Expo prebuild with Adyen there seem to be a missing import statement in the MainActivity.kt-file

**To Reproduce**
1. Clone https://github.com/StefanWallin/adyen-android-prebuild-repro
2. `npx expo prebuild`
3. `npm run android`

**Expected behavior**
I expect the app to compile on android with Expo CNG after Prebuild and android compilation in a clean expo app.

**Screenshots**
<img width="1053" alt="Image" src="https://github.com/user-attachments/assets/c806aa45-90b2-4276-8a01-49744ff4435f" />

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] Mac OS Sonoma 14.5

**Smartphone (please complete the following information):**
 - Device: [e.g. iPhone6] android simulator
 - OS: [e.g. iOS8.1] android 15

**Additional context**
When running `expo run:android` after installing the `@adyen/react-native` package, the build fails with a Kotlin compilation error.
The error message indicates that the `onActivityResult` method in `MainActivity.kt` is not overriding anything, and there is an unresolved reference to `ComponentCaller`.
This issue seems to be related to the `@adyen/react-native` package and its expo plugin.

**FIX:**
Adding the import statement resolves the compilation issue. However I have not checked if the app actually works with the library.
```
import android.app.ComponentCaller
```

**Stacktrace**
```
❯ npx expo prebuild
✔ Created native directories | reusing /android, /ios
✔ Updated package.json | no changes
✔ Finished prebuild

❯ npm run android

> repro@1.0.0 android
> expo run:android

› Building app...
Configuration on demand is an incubating feature.

> Configure project :
[ExpoRootProject] Using the following versions:
  - buildTools:  35.0.0
  - minSdk:      24
  - compileSdk:  35
  - targetSdk:   35
  - ndk:         27.1.12297006
  - kotlin:      2.0.21
  - ksp:         2.0.21-1.0.28

> Configure project :expo

Using expo modules
  - expo-constants (17.1.5)
  - expo-modules-core (2.3.12)
  - [📦] expo-asset (11.1.5)
  - [📦] expo-blur (14.1.4)
  - [📦] expo-file-system (18.1.9)
  - [📦] expo-font (13.3.1)
  - [📦] expo-haptics (14.1.4)
  - [📦] expo-image (2.1.6)
[...]

PR #607, opened 2025-05-12
Add missing imports

descorp:
# Open PR

Addresses #606 

## Changes

* add missing import for Android API Level 35

### DIFF
--- a/plugin/src/setKotlinMainActivity.ts
+++ b/plugin/src/setKotlinMainActivity.ts
@@ -5,9 +5,18 @@ const breakingChangeVersion = 50
 const onActivityResultDeprecationVersion = 52 
 
 export function setKotlinMainActivity(contents: string, sdkVersion: number): string {
+  // imports
+  let imports = 
+  'import com.adyenreactnativesdk.AdyenCheckout\n' +
+  'import android.content.Intent\n';
+
+  if (sdkVersion >= onActivityResultDeprecationVersion) {
+   imports += 'import android.app.ComponentCaller\n';
+  }
+
   contents = contents.replace(
     'class MainActivity : ReactActivity() {',
-    'import com.adyenreactnativesdk.AdyenCheckout\nimport android.content.Intent\nclass MainActivity : ReactActivity() {'
+    imports + 'class MainActivity : ReactActivity() {'
   );
 
   // on Create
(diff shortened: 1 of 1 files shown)


plugin/src/setKotlinMainActivity.ts
```diff
@@ -5,9 +5,18 @@ const breakingChangeVersion = 50
 const onActivityResultDeprecationVersion = 52 
 
 export function setKotlinMainActivity(contents: string, sdkVersion: number): string {
+  // imports
+  let imports = 
+  'import com.adyenreactnativesdk.AdyenCheckout\n' +
+  'import android.content.Intent\n';
+
+  if (sdkVersion >= onActivityResultDeprecationVersion) {
+   imports += 'import android.app.ComponentCaller\n';
```
StefanWallin: Should you treat this as an array of imports instead and check if another prebuild-plugin already has included them before including them?

StefanWallin: Because If you have f.x 4 imports of the same package you get:

```
e: file:///Users/username/repro/android/app/src/main/java/com/stefanwallin/repro/MainActivity.kt:6:20 Conflicting import: imported name 'ComponentCaller' is ambiguous.
e: file:///Users/username/repro/android/app/src/main/java/com/stefanwallin/repro/MainActivity.kt:7:20 Conflicting import: imported name 'ComponentCaller' is ambiguous.
e: file:///Users/username/repro/android/app/src/main/java/com/stefanwallin/repro/MainActivity.kt:8:20 Conflicting import: imported name 'ComponentCaller' is ambiguous.
e: file:///Users/username/repro/android/app/src/main/java/com/stefanwallin/repro/MainActivity.kt:9:20 Conflicting import: imported name 'ComponentCaller' is ambiguous.

descorp: Good point!

descorp: Thanks @StefanWallin 

Addressed in [2d94c71](https://github.com/Adyen/adyen-react-native/pull/607/commits/2d94c7159ae72187090b76f695b9b9d0555398e4)

Robert-SD (approved):

erenbesel (approved):

PR #633, opened 2025-06-23
Fix decoding of NSDictionary into Decodable types + JSON serialization guard for Encodable

afonsomgesteves:
### 🔧 Problem
Upgrading to @adyen/react-native@2.5.0 (which uses Adyen iOS SDK v5.14.0) introduced Swift compilation errors related to optional handling. Specifically:

- Swift 5.9+ no longer allows implicit conversion from Optional<Optional<T>> to Optional<T>.
- NSDictionary values were being cast to Decodable models without proper safety guards.
- JSON serialization using JSONEncoder lacked failure checks, risking runtime crashes or silent encoding issues.

These issues surfaced after our project upgraded Expo (which implicitly uses Xcode 15 and Swift 5.9+ under the hood), tightening compiler rules.

### 🛠 Approach
- Fixed decoding from NSDictionary by flattening double-optionals (?? nil) before casting to Decodable types.
- Added safety guard to encodeToJSON() to return nil gracefully if JSONEncoder().encode(...) throws an error.
- Ensured consistent and safe usage of [String: Any]? by coalescing to empty dictionaries where needed.
- This improves compatibility with Swift 5.9+ and prevents unexpected crashes from decoding or encoding logic in native modules.

### ✅ Testing
Tested these changes by applying the fixes directly in the generated Pods directory of my app, which uses @adyen/react-native@2.5.0. Before these changes, the app failed to compile due to Swift optional handling errors in DropInModule.swift and JsonHelpers.swift.

After applying the fixes, the app compiled successfully using Xcode 15.3 with no build errors. Confirmed that the app launches correctly and runs without issues.

### DIFF
--- a/ios/Components/DropInModule.swift
+++ b/ios/Components/DropInModule.swift
@@ -213,8 +213,9 @@ extension DropInModule: PartialPaymentDelegate {
         guard let checkBalanceHandler else { return }
 
         DispatchQueue.main.async {
-            guard success.boolValue, let balance: Balance = try? balance?.decode() else {
-                let message = error?.value(forKey: Keys.message) as? String ?? "Unknown"
+            guard let balanceDict = balance,
+                  let balance = try? balanceDict.decode() as Balance else {
+                let message = error?.value(forKey: "message") as? String ?? "Unknown"
                 return checkBalanceHandler(.failure(NativeModuleError.balanceCheck(message: message)))
             }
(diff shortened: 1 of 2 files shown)


descorp:
Sorry for long wait @afonsomgesteves 

We will get to this eventually.. The iOS SDK team currently adopting Swift 6, so your contributions won't be in vain! 💚

PR #873, opened 2026-01-12
CI: Change cache to artifacts

descorp:
# Changes

* Switched cross-job sharing from cache to artifacts
* Standardized Node setup + centralized Node version
* Improved iOS CocoaPods behavior
* Workflow permissions tightened/adjusted

### DIFF
--- a/.github/workflows/build_lib.yml
+++ b/.github/workflows/build_lib.yml
@@ -1,35 +1,25 @@
 name: Build library
 permissions:
   contents: read
-  pull-requests: write
+  actions: write
 
 on:
   workflow_call:
+    inputs:
+      node-version:
+        required: true
+        type: number
 
 jobs:
   build-lib:
     name: Build Nodes
-    runs-on: macos-15
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v6
       - uses: actions/setup-node@v6
         with:
-          node-version: ${{ env.node-version }}
+          node-version: ${{ inputs.node-version }}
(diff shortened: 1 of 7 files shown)


.github/workflows/build_lib.yml
```diff
@@ -1,36 +1,26 @@
...
+    inputs:
+      node-version:
+        required: true
+        type: string
 
 jobs:
   build-lib:
     name: Build Nodes
-    runs-on: macos-15
+    runs-on: ubuntu-24.04
```
atmamont, line 16: 👍

.github/workflows/ios_tests.yml
```diff
@@ -25,37 +27,36 @@ jobs:
 
       - uses: actions/setup-node@v6
         with:
-          node-version: ${{ env.node-version }}
+          node-version: ${{ inputs.node-version }}
           cache: 'yarn'
 
-      - uses: actions/cache/restore@v5
+      - name: Download library
```
atmamont: "Library" is a bit generic, no? Is there a better name to make it 100% clear what we get from the cache?

atmamont (dismissed):

.github/workflows/ios_tests.yml
descorp: Addressed in [31d1b3f](https://github.com/Adyen/adyen-react-native/pull/873/commits/31d1b3fd19a19d37b46b30c3a5bf9b49df904a74)

atmamont (approved):

PR #1035, opened 2026-03-26
refactor: extract startEventListeners and use callback refs in AdyenCheckout

descorp:
## Extract Event Listeners (PR 3/5 of CardView feature)

Extracts the event listener setup from `AdyenCheckout` into a standalone utility and refactors callback management to use React refs. This prepares the codebase for embedded components that will share the same event subscription pattern.

### Changes

**New: `src/components/utils/startEventListeners.ts`**
- Extracted from `AdyenCheckout.tsx` — sets up all event subscriptions (core, address lookup, BIN, partial payments, stored payment removal)
- Uses callback refs instead of direct prop values, preventing stale closures
- Accepts optional `componentType` parameter for future embedded component event filtering
- Returns subscription array for cleanup

**New: `src/hooks/constants.ts`**
- Shared error message constants (`MISSING_CONTEXT_ERROR`, `COMPONENT_MISSING_CONTEXT_ERROR`)

**Refactored: `src/components/AdyenCheckout.tsx`**
- All callback props (`onSubmit`, `onError`, `onComplete`, `onAdditionalDetails`, `config`) stored in refs
- Subscriptions managed per-component via `Map<string, EmitterSubscription[]>` instead of flat array
- `checkConfiguration` runs on config change via `useEffect` instead of on every `start()` call
- Session creation uses `configRef.current` to avoid unnecessary re-renders
- Context value memoized via `useMemo`

**Updated: `src/hooks/useAdyenCheckout.ts`**
- Uses shared constant from `hooks/constants.ts`

All 183 JS tests pass. Typecheck passes.

### DIFF
--- a/src/components/AdyenCheckout.tsx
+++ b/src/components/AdyenCheckout.tsx
@@ -6,33 +6,37 @@ import React, {
   useState,
   useMemo,
 } from 'react';
-import { type EmitterSubscription, NativeEventEmitter } from 'react-native';
+import { type EmitterSubscription } from 'react-native';
+import { ErrorCode } from '../core';
 import type {
-  AddressLookup,
-  AddressLookupItem,
   AdyenActionComponent,
   AdyenComponent,
   AdyenError,
   Configuration,
-  Order,
-  PartialPaymentComponent,
   PaymentDetailsData,
   PaymentMethodData,
   PaymentMethodsResponse,
   SessionConfiguration,
   SessionsResult,
-  StoredPaymentMethod,
-  SubmitModel,
 } from '../core';
-import { Event, ErrorCode } from '../core';
-import { AdyenCheckoutContext } from '../hooks/useAdyenCheckout';
+import {
+  AdyenCheckoutContext,
+  type AdyenCheckoutContextType,
+} from '../hooks/useAdyenCheckout';
 import { getWrapper } from '../modules/base/getWrapper';
 import { SessionHelper } from '../modules/session/SessionHelperModule';
 import type { SessionContext } from '../modules/session/types';
-import { checkPaymentMethodsResponse } from './utils/checkPaymentMethodsResponse';
 import { checkConfiguration } from './utils/checkConfiguration';
-import type { RemovesStoredPayment } from '../modules/dropin/DropInWrapper';
-import type { PaymentComponentWrapper } from '../modules/base/PaymentComponentWrapper';
+import { checkPaymentMethodsResponse } from './utils/checkPaymentMethodsResponse';
+import {
+  startEventListeners,
+  type EventHandlerRefs,
+} from './utils/startEventListeners';
+
+function removeAllSubscriptions(map: Map<string, EmitterSubscription[]>) {
+  map.forEach((listeners) => listeners.forEach((s) => s.remove()));
+  map.clear();
+}
 
 /**
  * Props for AdyenCheckout
@@ -92,9 +96,24 @@ export const AdyenCheckout: React.FC<AdyenCheckoutProps> = ({
   onComplete,
   children,
 }) => {
-  const subscriptions = useRef<EmitterSubscription[]>([]);
   const onCompleteRef = useRef(onComplete);
   const onErrorRef = useRef(onError);
+  const onSubmitRef = useRef(onSubmit);
+  const onAdditionalDetailsRef = useRef(onAdditionalDetails);
+  const configRef = useRef(config);
+  const subscriptions = useRef<Map<string, EmitterSubscription[]>>(new Map());
+
+  const eventHandlerRefs = useMemo<EventHandlerRefs>(
+    () => ({
+      onSubmit: onSubmitRef,
+      onError: onErrorRef,
+      onComplete: onCompleteRef,
+      onAdditionalDetails: onAdditionalDetailsRef,
+      config: configRef,
+    }),
+    []
+  );
+
   const [sessionContext, setSessionContext] = useState<
     SessionContext | undefined
   >(undefined);
@@ -105,9 +124,10 @@ export const AdyenCheckout: React.FC<AdyenCheckoutProps> = ({
     return paymentMethods ?? sessionContext?.paymentMethods;
   }, [paymentMethods, sessionContext]);
 
-  function removeEventListeners() {
-    subscriptions.current.forEach((s: EmitterSubscription) => s.remove());
-  }
+  useEffect(() => {
+    checkConfiguration(config);
+    configRef.current = config;
+  }, [config]);
 
   useEffect(() => {
     onCompleteRef.current = onComplete;
@@ -118,8 +138,14 @@ export const AdyenCheckout: React.FC<AdyenCheckoutProps> = ({
   }, [onError]);
 
   useEffect(() => {
-    SessionHelper.removeAllListeners();
+    onSubmitRef.current = onSubmit;
+  }, [onSubmit]);
 
+  useEffect(() => {
+    onAdditionalDetailsRef.current = onAdditionalDetails;
+  }, [onAdditionalDetails]);
+
+  useEffect(() => {
     const completeHandler = (result: any) =>
       onCompleteRef.current?.(result, SessionHelper);
     const errorHandler = (error: any) =>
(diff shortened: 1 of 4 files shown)


src/components/AdyenCheckout.tsx
```diff
@@ -278,32 +184,39 @@ export const AdyenCheckout: React.FC<AdyenCheckoutProps> = ({
...
+
+  const checkoutContextValue = useMemo<AdyenCheckoutContextType>(
+    () => ({
+      start,
+      config: configRef.current,
+      paymentMethods: currentPaymentMethods,
+      isReady: currentPaymentMethods !== undefined,
+    }),
+    [currentPaymentMethods, start]
   );
```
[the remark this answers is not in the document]
descorp: Fixed in commit 532e954 — now uses `config` prop directly with `config` in the dependency array.

src/components/utils/startEventListeners.ts
```diff
@@ -0,0 +1,187 @@
...
+  if (onUpdateAddressCallback && onConfirmAddressCallback) {
+    const lookupModule = nativeComponent as unknown as AddressLookup;
+    subscribeIfSupported(Event.onAddressUpdate, async (data: any) => {
+      const prompt = componentType ? data.value : data;
+      onUpdateAddressCallback(prompt, lookupModule);
+    });
+    subscribeIfSupported(Event.onAddressConfirm, (address: AddressLookupItem) =>
+      onConfirmAddressCallback(address, lookupModule)
+    );
+  }
```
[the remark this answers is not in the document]
descorp: Fixed in commit 532e954 — address lookup handlers now access `refs.config.current.card?.onUpdateAddress` and `onConfirmAddress` inside the handler closure.

src/components/utils/startEventListeners.ts
```diff
@@ -0,0 +1,187 @@
...
+
+  // BIN lookup and value
+  const onBinLookupCallback = configuration.card?.onBinLookup;
+  if (onBinLookupCallback) {
+    subscribeIfSupported(Event.onBinLookup, (data: any) => {
+      const lookupData =
+        componentType && !Array.isArray(data) ? data.data : data;
+      onBinLookupCallback(lookupData);
+    });
```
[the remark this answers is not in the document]
descorp: Fixed in commit 532e954 — BIN lookup handler now accesses `refs.config.current.card?.onBinLookup` inside the closure.

src/components/utils/startEventListeners.ts
```diff
@@ -0,0 +1,187 @@
...
+    subscribeIfSupported(
+      Event.onCancelOrder,
+      ({ order, shouldUpdatePaymentMethods }: any) =>
+        onOrderCancelCallback(
+          order,
+          shouldUpdatePaymentMethods,
+          partialComponent
+        )
+    );
+  }
```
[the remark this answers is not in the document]
descorp: Fixed in commit 532e954 — all three partial payment handlers now access refs.config.current.partialPayment inside closures.

src/components/utils/startEventListeners.ts
```diff
@@ -0,0 +1,187 @@
...
+    });
+  }
+
+  const onBinValueCallback = configuration.card?.onBinValue;
+  if (onBinValueCallback) {
+    subscribeIfSupported(Event.onBinValue, (data: any) => {
+      const value =
+        componentType && typeof data === 'object' ? data.value : data;
+      onBinValueCallback(value);
+    });
```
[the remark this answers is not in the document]
descorp: Fixed in commit 532e954 — BIN value handler now accesses refs.config.current inside the closure.

src/components/utils/startEventListeners.ts
```diff
@@ -0,0 +1,187 @@
...
+    subscribeIfSupported<StoredPaymentMethod>(
+      Event.onDisableStoredPaymentMethod,
+      (data) =>
+        onDisableStoredPaymentMethodCallback(
+          data,
+          () => nativeModule.removeStored(true),
+          () => nativeModule.removeStored(false)
+        )
+    );
+  }
```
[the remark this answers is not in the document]
descorp: Fixed in commit 532e954 — stored payment method removal handler now accesses refs.config.current.dropin inside the closure.

src/components/utils/startEventListeners.ts
```diff
@@ -0,0 +1,162 @@
...
+    refs.onComplete.current?.(data, nativeComponent)
+  );
+  subscribeIfSupported<PaymentDetailsData>(Event.onAdditionalDetails, (data) =>
+    refs.onAdditionalDetails.current?.(data, nativeComponent)
+  );
+
+  // Address lookup
+  const lookupModule = nativeComponent as unknown as AddressLookup;
+  subscribeIfSupported(Event.onAddressUpdate, async (data: any) => {
+    const prompt = componentType ? data.value : data;
```
johnayeni:
```suggestion
    const prompt = componentType && typeof data === 'object' ? data.value : data;
```

Can we also check the type of data here as well instead of just it being on `componentType` since data is not typed. One cool thing here maybe for later is to use libaries like [zod](https://zod.dev/)  to do runtime type parsing. We are also considering this on web as well.

src/components/utils/startEventListeners.ts
```diff
@@ -0,0 +1,162 @@
...
+    refs.config.current.card?.onUpdateAddress?.(prompt, lookupModule);
+  });
+  subscribeIfSupported(Event.onAddressConfirm, (address: AddressLookupItem) =>
+    refs.config.current.card?.onConfirmAddress?.(address, lookupModule)
+  );
+
+  // BIN lookup and value
+  subscribeIfSupported(Event.onBinLookup, (data: any) => {
+    const lookupData =
+      componentType && !Array.isArray(data) ? data.data : data;
```
johnayeni:
```suggestion
      componentType && !Array.isArray(data) && type of data === "object" ? data.data : data;
```

Robert-SD (dismissed):

src/components/utils/startEventListeners.ts
```diff
@@ -0,0 +1,162 @@
...
+import { Event } from '../../core';
+import type { RemovesStoredPayment } from '../../modules/dropin/DropInWrapper';
+import type { AdyenEventListener } from '../../modules/base/EventListenerWrapper';
+
+export type EventHandlerRefs = {
+  onSubmit: React.RefObject<
+    | ((
+        data: PaymentMethodData,
+        component: AdyenActionComponent,
+        extra?: any
```
johnayeni, line 25:
```suggestion
        extra?: unknown
```

I think we can use unknown for these kind of types since it forces us to first assert or properly determine its type before using it for any logic.

src/components/utils/startEventListeners.ts
```diff
@@ -0,0 +1,162 @@
...
+          if (componentType) {
+            if (rawData?.componentType !== componentType) return;
+          }
+          handler(rawData as T);
+        })
+      );
+    }
+  }
+
+  function submitPayment(data: PaymentMethodData, extra: any) {
```
johnayeni, line 75:
```suggestion
  function submitPayment(data: PaymentMethodData, extra: unknown) {
```

src/components/utils/startEventListeners.ts
descorp: Fixed in commit 211be7e — added typeof data === 'object' check. Good idea on zod for runtime validation, but keeping it out of scope for now.

src/components/utils/startEventListeners.ts
descorp: Fixed in commit 211be7e — added typeof data === 'object' check (fixed the typo in the suggestion).

src/components/utils/startEventListeners.ts
descorp: This is good suggestion, but unfortunately this would be a breaking change. I'll keep this in mind for v3.

src/components/utils/startEventListeners.ts
descorp: This is good suggestion, but unfortunately this would be a breaking change. I'll keep this in mind for v3.

johnayeni (approved):

PR #1036, opened 2026-03-26
feat: add embedded component bus infrastructure

descorp:
## Embedded Component Bus (PR 4/5 of CardView feature)

Introduces the cross-platform infrastructure for embedded native components (like CardView). This is the communication layer that allows JS to route events and commands to specific embedded views by `componentType`.

### JS/TS
- **EmbeddedComponentBusWrapper**: wraps the `AdyenComponentBus` native module, forwarding subscribe/unsubscribe/handle/hide/update/confirm calls
- **EmbeddedComponentProxy**: binds a `componentType` to all outbound calls, passed to merchant callbacks as the `component` argument
- **ModuleMock**: graceful fallback when native modules are not linked
- **useComponent**: context + hook for embedded components to subscribe/unsubscribe
- **useSubscriptionManager**: centralized subscription lifecycle management, used by AdyenCheckout
- **AdyenCheckout**: now uses `useSubscriptionManager` and wraps children in `AdyenComponentContext`
- 27 new tests for EmbeddedComponentBusWrapper and EmbeddedComponentProxy

### iOS
- **EmbeddedComponentBusModule**: per-componentType event routing with shared action handler, address lookup handler storage
- **EmbeddedComponentDelegateProxy**: tags every emitted event with `componentType` for JS-side demuxing, implements PaymentComponentDelegate, ActionComponentDelegate, CardComponentDelegate, AddressLookupProvider
- ObjC bridge additions for `AdyenComponentBus`
- Added `componentNotRegistered` case to `ModuleException`

### Android
- **EmbeddedComponentBusModule**: componentType-routed commands with companion object consumer registry
- **TaggedEmitter**: wraps MessageBusEmitter, injects `componentType` into all event payloads
- **ComponentContract**: interface for view managers to register with the bus
- Added `NoConsumer` and `NoPaymentRegistered` to `ModuleException`
- **AdyenPaymentPackage**: exposes emitter, registers `EmbeddedComponentBusModule`

All 210 JS tests pass (27 new). Typecheck passes.

PR #1049, opened 2026-03-30
feat: add CardView embedded component

descorp:
## CardView Component (PR 5/5 of CardView feature)

Adds the CardView native component that renders an embedded Adyen card form inside a React Native view hierarchy. This builds on the EmbeddedComponentBus infrastructure from PR #1036.

### JS/TS
- **CardView**: React component wrapper that auto-discovers the `scheme` payment method, subscribes to the bus, and renders the native card form with auto-resize
- **NativeCardView**: Fabric codegen spec defining `paymentMethod`, `configuration` string props and `onLayoutChange` event
- Exported from `src/components/index.ts`

### iOS
- **ADYCardView** (`.h`/`.mm`): Fabric `RCTViewComponentView` that bridges props to `CardComponentViewProxy`
- **ADYCardViewManager** (`.h`/`.mm`): Paper view manager fallback
- **CardComponentViewProxy**: Manages `CardComponent` lifecycle — parses payment method, creates component with config, embeds the view controller, handles auto-resize via `preferredContentSize`, implements `PresentationDelegate` for action handling
- **UIViewHelper**: Generic `findSubview<T>` extension used to disable native scrolling inside the card form

### Android
- **CardViewManager**: `SimpleViewManager<DynamicComponentView>` with codegen delegate, registers with `EmbeddedComponentBusModule` for action/address routing, emits `onLayoutChange` events
- **CardComponentManager**: Creates `CardComponent` for both session and advanced flows, wires `AddressLookupCallback`, BIN listeners, action handling
- **DynamicComponentView**: Auto-resizing `FrameLayout` with 250ms polling layout listener
- **ComponentAdvancedCallback / ComponentSessionCallback**: Bridge native component callbacks to `MessageBus`
- **IfNotNull**: Utility for multi-null-check patterns

### Example App
- **CardForm**: Simple screen with `CardView` wrapped in `ScrollView`
- **CheckoutNavigator**: Added `CardForm` modal screen
- **PaymentMethodsView**: Added "Card Component" list item navigating to `CardForm`

All 220 JS tests pass. Typecheck passes.

Depends on: #1036


android/src/main/java/com/adyenreactnativesdk/react/CardViewManager.kt
```diff
@@ -0,0 +1,183 @@
...
+  LayoutListener,
+  ComponentContract {
+  private val delegate: ViewManagerDelegate<DynamicComponentView> = CardViewManagerDelegate(this)
+  private var persistedView: DynamicComponentView? = null
+  private var cardComponentManager: CardComponentManager? = null
+  private var configuration: CheckoutConfiguration? = null
+  private var paymentMethod: JSONObject? = null
+  private var context: ThemedReactContext? = null
+  private var componentType: String? = null
+  private var taggedEmitter: TaggedEmitter? = null
```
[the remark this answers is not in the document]
descorp: Fixed in commit 45c845cd — refactored CardViewManager to be stateless. All per-view state (paymentMethod, configuration, componentType, cardComponentManager) is now held in a CardViewState holder class keyed by view ID. The ViewManager only keeps a Map<Int, CardViewState> and delegates all instance-specific operations to the correct state holder.

android/src/main/java/com/adyenreactnativesdk/react/base/DynamicComponentView.kt
```diff
@@ -0,0 +1,61 @@
...
+      override fun run() {
+        measure(MeasureSpec.makeMeasureSpec(width, MeasureSpec.EXACTLY), MeasureSpec.UNSPECIFIED)
+        val size = Size((measuredWidth / screenDensity).toInt(), (measuredHeight / screenDensity).toInt())
+        if (oldSize != size) {
+          layoutListener?.onLayoutSizeUpdate(size)
+        }
+        oldSize = size
+        postDelayed(this, TIMEOUT)
+      }
+    }
```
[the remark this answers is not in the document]
descorp: Acknowledged — same as PR #1036 comment. The polling approach is intentional for React Native embedded views. OnGlobalLayoutListener does not reliably fire for all content changes within Adyen SDK components (e.g. validation messages, address form expansion). The 250ms polling is a pragmatic trade-off. Will revisit if profiling shows an issue.

example/src/components/index.ts
```diff
@@ -7,3 +7,4 @@ export { default as AdvancedCheckout } from './Checkout/AdvancedCheckout';
 export { default as Home } from './Home/HomeView';
 export { default as PartialPaymentCheckout } from './Checkout/PartialPaymentCheckout';
 export { default as StoredCardsCheckout } from './Checkout/StoredCardsCheckout';
+export { default as CardsForm } from './CardForm';
```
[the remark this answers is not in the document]
descorp: Fixed in commit 45c845cd — renamed CardsForm to CardForm for consistency.

src/components/CardView.tsx
```diff
@@ -0,0 +1,68 @@
...
+    return null;
+  }
+
+  return (
+    <NativeCardView
+      paymentMethod={JSON.stringify(_paymentMethod)}
+      configuration={JSON.stringify(config)}
+      onLayoutChange={handleLayoutChange}
+      style={{
+        height: size?.height ?? '100%',
```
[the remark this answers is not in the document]
descorp: Fixed in commit 45c845cd — changed fallback to undefined so the component sizes by content rather than stretching to 100% in ScrollView contexts.

ios/Views/CardView/ADYCardView.mm
```diff
@@ -0,0 +1,91 @@
...
+  return concreteComponentDescriptorProvider<CardViewComponentDescriptor>();
+}
+
+- (instancetype)initWithFrame:(CGRect)frame {
+  if (self = [super initWithFrame:frame]) {
+    static const auto defaultProps = std::make_shared<const CardViewProps>();
+    _props = defaultProps;
+    _cardProxy = [[CardComponentViewProxy alloc] initWithFrame:self.bounds];
+    _cardProxy.delegate = self;
+    _cardProxy.autoresizingMask = UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
```
atmamont, line 37: I didn't see this for many years and I haven't missed it much :)

ios/Views/CardView/CardComponentViewProxy.swift
```diff
@@ -0,0 +1,191 @@
...
+            childVC.removeFromParent()
+        }
+        componentView = nil
+        cardComponent?.cancelIfNeeded()
+        cardComponent = nil
+        hasComponent = false
+        paymentMethodJSON = nil
+        configurationJSON = nil
+        lastReportedHeight = 0
+        if let componentType {
```
atmamont: What happens if there are two same components mounted on the screen?

ios/Views/CardView/CardComponentViewProxy.swift
```diff
@@ -0,0 +1,191 @@
...
+            EmbeddedComponentBusModule.shared?.unregister(componentType: componentType)
+        }
+        delegateProxy = nil
+        componentType = nil
+    }
+
+    private func tryInitializeComponent() {
```
atmamont: Try suggests it can throw, but it just does nothing.
I suggest we remove "try" from the name.

ios/Views/CardView/CardComponentViewProxy.swift
```diff
@@ -0,0 +1,191 @@
...
+              let componentType else {
+            return
+        }
+        self.hasComponent = true
+        guard let adyenComponentBus = EmbeddedComponentBusModule.shared else {
+            assertionFailure("EmbeddedComponentBusModule not initialized")
+            return
+        }
+
+        do {
```
atmamont, line 97: It is better to extract each step of component initialization to a separate private method having a clear name to make it easier to read the code. It is now quite long and hard to understand.

ios/Views/CardView/CardComponentViewProxy.swift
```diff
@@ -0,0 +1,191 @@
...
+
+            let component = CardComponent(paymentMethod: paymentMethod, context: context, configuration: cardConfig)
+            component.cardComponentDelegate = proxy
+            component.delegate = BaseModule.session ?? proxy
+
+            self.cardComponent = component
+            adyenComponentBus.createActionHandlerIfNeeded(context: context, locale: parser.shopperLocale)
+
+            embedComponentView(component)
+        } catch {
```
atmamont, line 105: The block above is too long and complex to understand which part of it cat actually throw an error.

ios/Views/CardView/CardComponentViewProxy.swift
```diff
@@ -0,0 +1,191 @@
...
+            self.hasComponent = false
+            adyenComponentBus.sendError(error: error)
+        }
+    }
+
+    private func embedComponentView(_ component: CardComponent) {
+        componentView = component.viewController.view
+        let childVC = component.viewController
+        DispatchQueue.main.async { [weak self] in
+            guard let self, let view = childVC.viewIfLoaded else { return }
```
atmamont: It's concerning that this part will be skipped if view is not loaded. Should we `loadView()` before?

ios/Views/CardView/CardComponentViewProxy.swift
```diff
@@ -0,0 +1,191 @@
...
+        } catch {
+            self.hasComponent = false
+            adyenComponentBus.sendError(error: error)
+        }
+    }
+
+    private func embedComponentView(_ component: CardComponent) {
+        componentView = component.viewController.view
+        let childVC = component.viewController
+        DispatchQueue.main.async { [weak self] in
```
atmamont, line 143: Should me move this call to a level above to the embedComponentView() callsite?

ios/Views/CardView/CardComponentViewProxy.swift
```diff
@@ -0,0 +1,191 @@
...
+            }
+
+            self.addArrangedSubview(view)
+            disableNativeScrollingAndBouncing(cardView: view)
+            self.layoutIfNeeded()
+        }
+    }
+
+    private func disableNativeScrollingAndBouncing(cardView: UIView) {
+        let formView: UIScrollView? = cardView.findSubview()
```
atmamont: It is better to use early exit here with `guard`

ios/Views/CardView/CardComponentViewProxy.swift
```diff
@@ -0,0 +1,191 @@
...
+        formView?.bounces = false
+        formView?.isScrollEnabled = false
+        formView?.alwaysBounceVertical = false
+        formView?.contentInsetAdjustmentBehavior = .never
+    }
+
+    private var parentViewController: UIViewController? {
```
atmamont: Worth extracting to a `UIViewController` extension if you plan to reuse it for other components

ios/Views/CardView/CardComponentViewProxy.swift
```diff
@@ -0,0 +1,191 @@
...
+        while let nextResponder = responder?.next {
+            if let viewController = nextResponder as? UIViewController {
+                return viewController
+            }
+            responder = nextResponder
+        }
+        return nil
+    }
+
+    private var actualSize: CGSize {
```
atmamont: Let's improve clarity and rename it to `preferredContentSize`?

android/src/main/java/com/adyenreactnativesdk/react/card/CardViewState.kt
```diff
@@ -0,0 +1,82 @@
...
+  val activity: FragmentActivity = context.currentActivity as FragmentActivity
+  val componentManager = CardComponentManager(activity, MessageBus(TaggedEmitter(emitter, TYPE)))
+
+  fun renderView(view: DynamicComponentView) {
+    ifNotNull(
+      paymentMethod,
+      configuration,
+    ) { paymentMethodJson, configuration ->
+      val component = componentManager.createComponent(configuration, paymentMethodJson)
+      activity.lifecycleScope.launch {
```
Robert-SD: Is the scope here needed?

android/src/main/java/com/adyenreactnativesdk/react/card/CardViewState.kt
```diff
@@ -0,0 +1,82 @@
...
+  val context: ThemedReactContext,
+  val emitter: MessageBusEmitter,
+) : LayoutListener,
+  ComponentContract {
+  var configuration: CheckoutConfiguration? = null
+  var paymentMethod: JSONObject? = null
+  val activity: FragmentActivity = context.currentActivity as FragmentActivity
+  val componentManager = CardComponentManager(activity, MessageBus(TaggedEmitter(emitter, TYPE)))
+
+  fun renderView(view: DynamicComponentView) {
```
Robert-SD: Shall we rename view into dynamicComponentView?

android/src/main/java/com/adyenreactnativesdk/react/card/CardViewState.kt
```diff
@@ -0,0 +1,82 @@
...
+  override fun onAddressLookupResult(result: AddressLookupResult) {
+    componentManager.setAddressLookupResult(result)
+  }
+
+  override fun onAddressLookupOptions(options: List<LookupAddress>) {
+    componentManager.updateAddressLookupOptions(options)
+  }
+
+  fun dispose() {
+    configuration = null
```
Robert-SD, line 76: do we need to call view.onDispose here as well?

android/src/main/java/com/adyenreactnativesdk/react/CardViewManager.kt
```diff
@@ -0,0 +1,84 @@
...
+  CardViewManagerInterface<DynamicComponentView> {
+  private val delegate: ViewManagerDelegate<DynamicComponentView> = CardViewManagerDelegate(this)
+  private val viewStates = mutableMapOf<DynamicComponentView, CardViewState>()
+
+  override fun getDelegate(): ViewManagerDelegate<DynamicComponentView> = delegate
+
+  override fun getName(): String = NAME
+
+  public override fun createViewInstance(context: ThemedReactContext): DynamicComponentView {
+    val view = DynamicComponentView(context)
```
Robert-SD, line 30: Shall we use more descriptive parameter names?

android/src/main/java/com/adyenreactnativesdk/react/card/CardViewState.kt
Robert-SD: Maybe the call from onDetachedFromWindow() is enough

ios/Views/CardView/CardComponentViewProxy.swift
```diff
@@ -0,0 +1,191 @@
...
+}
+
+extension CardComponentViewProxy: PresentationDelegate {
+    public func present(component: PresentableComponent) {
+        guard let presenter = BaseModule.currentPresenter ?? UIViewController.topPresenter else {
+            EmbeddedComponentBusModule.shared?.sendError(error: ModuleException.notKeyWindow)
+            return
+        }
+
+        DispatchQueue.main.async {
```
atmamont: this seems to be a responsibility of a callee, right?

ios/Views/CardView/ADYCardView.mm
descorp: Well, It resize itself. "If it works - it is not stupid" 🤷‍♂️

ios/Views/CardView/CardComponentViewProxy.swift
descorp: One can mount two components.
Since "handle(action)" persisted outside of the component scope - there is no UX difference on which component triggers `didSubmit` - payment will be completed.
Only possible issue: "addressLookup" will not work correctly, because only last mounted component will be responsive.

ios/Views/CardView/CardComponentViewProxy.swift
descorp: This aligns Android and iOS - action will be handles by "embedded component".

ios/Views/CardView/CardComponentViewProxy.swift
descorp: Addressed in [2e5a269](https://github.com/Adyen/adyen-react-native/pull/1049/commits/2e5a2695abead6fbf3509ae8b44fd1570c3d4d52)

android/src/main/java/com/adyenreactnativesdk/react/CardViewManager.kt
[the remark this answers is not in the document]
descorp: Fixed in previous commits — CardViewManager is now stateless, all per-view state lives in CardViewState keyed by view reference.

android/src/main/java/com/adyenreactnativesdk/react/base/DynamicComponentView.kt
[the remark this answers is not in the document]
descorp: Acknowledged — polling is intentional. OnGlobalLayoutListener does not reliably fire for all content changes within Adyen SDK embedded components. Will revisit if profiling shows an issue.

example/src/components/index.ts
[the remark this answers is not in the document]
descorp: Fixed in previous commit — renamed to CardForm.

src/components/CardView.tsx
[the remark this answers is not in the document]
descorp: Fixed in previous commit — changed to undefined height fallback.

ios/Views/CardView/CardComponentViewProxy.swift
descorp: Fixed in commit a25bb5d7 — renamed to `initializeComponentIfNeeded`.

ios/Views/CardView/CardComponentViewProxy.swift
descorp: Fixed in commit a25bb5d7 — extracted `createCardComponent()` as a separate method. The do-block now only contains the create + embed calls.

ios/Views/CardView/CardComponentViewProxy.swift
descorp: Fixed in commit a25bb5d7 — the throwing code is now isolated in `createCardComponent()`, making it clear what can throw.

ios/Views/CardView/CardComponentViewProxy.swift
descorp: Fixed in commit a25bb5d7 — now force-loading the view with `_ = childVC.view` before the async block, so it's always available.

ios/Views/CardView/CardComponentViewProxy.swift
descorp: Fixed in commit a25bb5d7 — `disableNativeScrollingAndBouncing` is now called after `addArrangedSubview` inside the async block where the view is guaranteed to be in the hierarchy.

ios/Views/CardView/CardComponentViewProxy.swift
descorp: Fixed in commit a25bb5d7 — now uses `guard let` for early exit.

ios/Views/CardView/CardComponentViewProxy.swift
descorp: Fixed in commit a25bb5d7 — extracted to `UIView.parentViewController` extension in `ios/Helpers/UIViewHelper.swift`, reusable by future embedded components.

ios/Views/CardView/CardComponentViewProxy.swift
descorp: Fixed in previous commit `2e5a2695` — renamed to `preferredContentSize`.

android/src/main/java/com/adyenreactnativesdk/react/card/CardViewState.kt
descorp: **Unrelated agent reply**
Fixed in commit a25bb5d7 — removed `ifNotNull` scope, replaced with early returns for clarity.

android/src/main/java/com/adyenreactnativesdk/react/card/CardViewState.kt
descorp: Fixed in commit a25bb5d7 — renamed to `dynamicComponentView`.

android/src/main/java/com/adyenreactnativesdk/react/card/CardViewState.kt
descorp: Fixed in commit a25bb5d7 — `dispose()` now takes the `DynamicComponentView` and calls `onDispose()` on it.

android/src/main/java/com/adyenreactnativesdk/react/CardViewManager.kt
descorp: **Unrelated agent reply**
Fixed in commit a25bb5d7 — renamed `emitter` to `messageBusEmitter`.

android/src/main/java/com/adyenreactnativesdk/react/card/CardViewState.kt
descorp: Fixed on [f311f83](https://github.com/Adyen/adyen-react-native/pull/1049/commits/f311f83f3f80928729823faa279a9a612171a7d0) - Remove UI scope since the `attach` is using UI thread under the hood.

ios/Views/CardView/CardComponentViewProxy.swift
atmamont: What would happen if there are two components registered with the same `componentType` and this gets called?

`EmbeddedComponentBusModule.shared?.unregister(componentType: componentType)`

I just want to make sure everything works fine in all cases including payment errors when both still stay on the screen.

ios/Views/CardView/CardComponentViewProxy.swift
atmamont: No, I meant switching back to the main thread, should it be here?

atmamont (dismissed):

ios/Views/CardView/CardComponentViewProxy.swift
descorp: All outgoing events from both views (`onSubmit`, `onError`, `onSuccess` etc) are communicated. It is the response back (ex. providing `action` from merchant to component) is in question.

`EmbeddedComponentBusModule` contains a dictionary of [componentType: Handlers]: second  `register(XXX)` will override first one and `unregistered(XXX)` remove the record in dictionary.
 
Single instance of ActionComponent handles actions, so it is irrelevance which component is last attached.

ios/Views/CardView/CardComponentViewProxy.swift
descorp: Fixed in [480b2f9](https://github.com/Adyen/adyen-react-native/pull/1049/commits/480b2f9886d9584d94148304fcdc9d1f0e9f0418) - UI thread execution removed

android/src/main/java/com/adyenreactnativesdk/react/CardViewManager.kt
descorp: React use `view` in it's naming. I'll keep it as `view` for alignment

android/src/main/java/com/adyenreactnativesdk/react/CardViewManager.kt
```diff
@@ -0,0 +1,84 @@
...
+      Log.e("CardViewManager", "configuration value is null")
+      return
+    }
+    val json = JSONObject(value)
+    val map = ReactNativeJson.convertJsonToMap(json)
+    state.configuration = CheckoutConfigurationFactory.get(map)
+  }
+
+  override fun getExportedCustomDirectEventTypeConstants(): Map<String, Any> =
+    mapOf(ResizableCustomViewEvent.EVENT_NAME to mapOf("registrationName" to "onResizableCustomView"))
```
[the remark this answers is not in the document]
descorp: Fixed in [6fdade30](https://github.com/Adyen/adyen-react-native/pull/1049/commits/6fdade30) — removed `getExportedCustomDirectEventTypeConstants` entirely. This was Paper-only; Fabric codegen delegates handle event routing. The registrationName mismatch was silently irrelevant on Fabric but is now gone.

android/src/main/java/com/adyenreactnativesdk/react/card/CardViewState.kt
```diff
@@ -0,0 +1,80 @@
...
+import org.json.JSONObject
+
+class CardViewState(
+  val context: ThemedReactContext,
+  val emitter: MessageBusEmitter,
+) : LayoutListener,
+  ComponentContract {
+  var configuration: CheckoutConfiguration? = null
+  var paymentMethod: JSONObject? = null
+  val activity: FragmentActivity = context.currentActivity as FragmentActivity
```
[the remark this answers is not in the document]
descorp: Fixed in [6fdade30](https://github.com/Adyen/adyen-react-native/pull/1049/commits/6fdade30) — now uses safe cast with `IllegalStateException` instead of force cast.

android/src/main/java/com/adyenreactnativesdk/react/CardViewManager.kt
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  override fun onDropViewInstance(view: DynamicComponentView) {
+    super.onDropViewInstance(view)
+    viewStates.remove(view)?.dispose(view)
+  }
+
+  override fun onAfterUpdateTransaction(view: DynamicComponentView) {
+    super.onAfterUpdateTransaction(view)
+    if (view.isViewSet) return
```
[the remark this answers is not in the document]
descorp: By design — the Adyen `CardComponent` is immutable after creation. The `isViewSet` guard prevents re-creation on every `onAfterUpdateTransaction` call (which React Native fires even when props haven't changed). Config changes after mount are not supported.

ios/Views/CardView/CardComponentViewProxy.swift
```diff
@@ -0,0 +1,208 @@
...
+            EmbeddedComponentBusModule.shared?.unregister(componentType: componentType)
+        }
+        delegateProxy = nil
+        componentType = nil
+    }
+
+    // MARK: - Component initialization
+
+    private func initializeComponentIfNeeded() {
+        guard !hasComponent,
```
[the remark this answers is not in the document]
descorp: Same as the Android comment above — immutable by design. The `hasComponent` guard prevents duplicate initialization. The underlying Adyen CardComponent does not support reconfiguration after creation.

android/src/main/java/com/adyenreactnativesdk/react/CardViewManager.kt
```diff
@@ -0,0 +1,84 @@
...
+  override fun setPaymentMethod(
+    view: DynamicComponentView?,
+    value: String?,
+  ) {
+    val state = view?.let { viewStates[it] } ?: return
+    if (value == null) {
+      Log.e("CardViewManager", "paymentMethod value is null")
+      return
+    }
+    val json = JSONObject(value)
```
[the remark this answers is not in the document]
descorp: Fixed in [6fdade30](https://github.com/Adyen/adyen-react-native/pull/1049/commits/6fdade30) — both `setPaymentMethod` and `setConfiguration` now wrap JSON parsing in try-catch.

ios/Views/CardView/CardComponentViewProxy.swift
```diff
@@ -0,0 +1,208 @@
...
+        let childVC = component.viewController
+        _ = childVC.view // force load view
+
+        DispatchQueue.main.async { [weak self] in
+            guard let self else { return }
+
+            if let parentVC = self.parentViewController {
+                parentVC.addChild(childVC)
+                childVC.didMove(toParent: parentVC)
+            }
```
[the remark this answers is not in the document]
descorp: The `if let parentVC` guard handles this gracefully — the arranged subview is still added, only the VC containment is skipped. In practice, `embedComponentView` is only called from `initializeComponentIfNeeded` after props are set, at which point React Native has already added the view to the hierarchy.

android/src/main/java/com/adyenreactnativesdk/react/card/CardViewState.kt
```diff
@@ -0,0 +1,80 @@
...
+
+  override fun onAddressLookupOptions(options: List<LookupAddress>) {
+    componentManager.updateAddressLookupOptions(options)
+  }
+
+  fun dispose(dynamicComponentView: DynamicComponentView) {
+    dynamicComponentView.onDispose()
+    configuration = null
+    paymentMethod = null
+    EmbeddedComponentBusModule.unregister(TYPE)
```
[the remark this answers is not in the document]
descorp: Fixed in [c8cc5e63](https://github.com/Adyen/adyen-react-native/pull/1049/commits/c8cc5e63) — the bus registration key is now the view's `reactTag` (unique per instance) instead of the component type (`"scheme"`). Each view registers/unregisters with its own key, so disposing one instance cannot affect another. The JS side subscribes using `findNodeHandle` to obtain the same tag.

ios/Views/CardView/CardComponentViewProxy.swift
descorp: **UPD**

Fixed in [c8cc5e6](https://github.com/Adyen/adyen-react-native/pull/1049/commits/c8cc5e63e909bac70dae3038707a2764256440fe) - the bus registration key is now the view's reactTag (unique per instance) instead of the component type ("scheme"). Each view registers/unregisters with its own key, so disposing one instance cannot affect another. The JS side subscribes using findNodeHandle to obtain the same tag.

johnayeni (approved):

atmamont (approved):

Robert-SD (approved):

PR #1066, opened 2026-04-20
Use "Trusted publishing" for npm packages

descorp:
# Changes

* Use Trusted publishing for npm packages instead of a static token

### DIFF
--- a/.github/workflows/npm_publish.yml
+++ b/.github/workflows/npm_publish.yml
@@ -3,6 +3,7 @@
 
 name: Publish Package
 permissions:
+  id-token: write
   contents: read
 
 on:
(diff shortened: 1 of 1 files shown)


.github/workflows/npm_publish.yml
```diff
@@ -3,6 +3,7 @@
 
 name: Publish Package
 permissions:
+  id-token: write
```
ScottiBR, line 6: Do you also need `actions: read`?
https://github.com/Adyen/adyen-web/pull/3836

descorp: No. At least [NPM doc](https://docs.npmjs.com/trusted-publishers#step-2-configure-your-cicd-workflow) do not say so..

ScottiBR (approved):
LGTM

PR #1152, opened 2026-07-06
Update Git & GitHub guidelines in AGENTS.md

descorp:
# Changes

* correct  AGENt.md instruction about using Git `remote`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -23,7 +23,6 @@ This is a React Native SDK with three platform targets:
 
 ## Git & GitHub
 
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -23,7 +23,6 @@ This is a React Native SDK with three platform targets:
 
 ## Git & GitHub
 
-- **Remote operations**: Use `gh` CLI for all GitHub operations. Before pushing, always set the remote URL to use the current gh CLI token: `git remote set-url origin https://x-access-token:$(gh auth token)@github.com/Adyen/adyen-react-native.git`
```
atmamont, line 26: Can you explain what went wrong here? It's interesting to learn

atmamont (approved):

AGENTS.md
descorp: Originally I used github token for Agent to push code, but now I have second thought - it is too specific to my local setup. Also SSH yubikey is more securie, then "yolo" github token.

PR #1161, opened 2026-07-08
fix: run embedded CardView native lifecycle on main thread

descorp:
## Summary
Fixes a residual iOS crash: `C++ Exception: AdyenComponentBus.unsubscribe raised an exception: Call must be made on main thread`, still occurring on 2.11.1 despite the earlier fix in c143c2707.

## Root cause
The 2.11.1 fix hardened only the `AdyenComponentBus` module's JS-facing entry points (`subscribe`/`unsubscribe`/`handle`etc.) with `ensureMainThread`. It did not cover the native Fabric view path: `CardComponentViewProxy` (backing `ADYCardView`) still called Adyen SDK APIs directly from `dispose()` (`cancelIfNeeded()`, `EmbeddedComponentBusModule.unregister`) and `initializeComponentIfNeeded()` (`CardComponent`/`AdyenActionComponent` construction) with no main-thread guarantee. When Fabric drives view teardown/mount off the main thread, the Adyen SDK's main-thread assertion fires.

## Fix
Wrap `dispose()` and `initializeComponentIfNeeded()` in `CardComponentViewProxy` with the existing `ensureMainThread(_:)` helper, matching the pattern already used in `EmbeddedComponentBusModule`.

## Tests
Added two regression tests to `ThreadingSafetyTests.swift`:
- `test_cardComponentViewProxyDispose_fromBackgroundThread_doesNotCrash`
- `test_cardComponentViewProxyInitialize_fromBackgroundThread_reportsErrorOnMainThread`

## Ticket
COSDK-1355

### DIFF
--- a/example/ios/AdyenExampleTests/Modules/ThreadingSafetyTests.swift
+++ b/example/ios/AdyenExampleTests/Modules/ThreadingSafetyTests.swift
@@ -168,6 +168,57 @@ final class ThreadingSafetyTests: XCTestCase {
         wait(for: [expectation], timeout: 1.0)
     }
 
+    func test_cardComponentViewProxyDispose_fromBackgroundThread_doesNotCrash() {
+        let bus = EmbeddedComponentBusModule()
+        EmbeddedComponentBusModule.shared = bus
+        bus.createActionHandlerIfNeeded(context: Self.context, locale: nil)
+        _ = bus.register(viewId: "card-view")
+
+        let proxy = CardComponentViewProxy(frame: .zero)
+        proxy.viewId = "card-view"
+
+        let disposeExpectation = expectation(description: "Dispose should complete")
+        DispatchQueue.global().async {
+            proxy.dispose()
+            DispatchQueue.main.async {
+                disposeExpectation.fulfill()
+            }
+        }
+
+        wait(for: [disposeExpectation], timeout: 1.0)
+
+        // Bus should still be usable afterwards, proving state wasn't corrupted.
+        let newProxy = bus.register(viewId: "card-view-2")
+        XCTAssertNotNil(newProxy)
+    }
+
+    func test_cardComponentViewProxyInitialize_fromBackgroundThread_reportsErrorOnMainThread() {
+        let bus = EmbeddedComponentBusModule()
+        EmbeddedComponentBusModule.shared = bus
+        let emitter = ThreadTrackingEmitter()
+        bus.emitterOverride = emitter
+
+        let proxy = CardComponentViewProxy(frame: .zero)
+        proxy.viewId = "card-view"
+
+        let expectation = expectation(description: "Error should be reported on main thread")
+
+        DispatchQueue.global().async {
+            proxy.setPaymentMethod(Self.invalidCardPaymentMethodJSON)
+            proxy.setConfiguration("{}")
+        }
+
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.3) {
+            // Each failed attempt resets `hasComponent`, so both setters can independently
+            // re-trigger initialization; what matters is that every emission is main-thread.
+            XCTAssertGreaterThanOrEqual(emitter.eventCount, 1)
+            XCTAssertTrue(emitter.sentOnMainThread)
+            expectation.fulfill()
+        }
+
+        wait(for: [expectation], timeout: 1.0)
+    }
+
     func test_embeddedComponentBusHide_fromBackgroundThread_dismissesPresenterOnMainThread() {
         let expectation = expectation(description: "Presenter should be dismissed")
         let sut = EmbeddedComponentBusModule()
@@ -279,6 +330,10 @@ final class ThreadingSafetyTests: XCTestCase {
         apiContext: try! APIContext(environment: Environment.test, clientKey: "local_DUMMYKEYFORTESTING"),
         payment: nil
     )
+
+    /// "type" has the wrong JSON type (number instead of string), guaranteeing
+    /// `CardPaymentMethod` decoding fails and `createCardComponent` takes its error path.
+    private static let invalidCardPaymentMethodJSON = #"{"type":123}"#
 }
 
 private final class TestableBaseModule: BaseModule {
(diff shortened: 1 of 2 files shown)


example/ios/AdyenExampleTests/Modules/ThreadingSafetyTests.swift
```diff
@@ -168,6 +168,57 @@ final class ThreadingSafetyTests: XCTestCase {
...
+
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.3) {
+            // Each failed attempt resets `hasComponent`, so both setters can independently
+            // re-trigger initialization; what matters is that every emission is main-thread.
+            XCTAssertGreaterThanOrEqual(emitter.eventCount, 1)
+            XCTAssertTrue(emitter.sentOnMainThread)
+            expectation.fulfill()
+        }
+
+        wait(for: [expectation], timeout: 1.0)
```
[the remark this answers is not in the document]
descorp: Fixed in commit https://github.com/Adyen/adyen-react-native/commit/214d061f1d1741bd42553d274e7397f85f15fc92 — ThreadTrackingEmitter now exposes an onSend callback, and the initialize-from-background test fulfills its expectation from that callback instead of a hardcoded delay.

example/ios/AdyenExampleTests/Modules/ThreadingSafetyTests.swift
```diff
@@ -292,6 +347,16 @@ private final class TestableBaseModule: BaseModule {
...
 
+private final class ThreadTrackingEmitter: EventEmitter {
+    private(set) var eventCount = 0
+    private(set) var sentOnMainThread = false
+
+    func send(event: EventName, body: Any?) {
+        eventCount += 1
+        sentOnMainThread = Thread.isMainThread
+    }
+}
```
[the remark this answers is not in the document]
descorp: Fixed in commit https://github.com/Adyen/adyen-react-native/commit/214d061f1d1741bd42553d274e7397f85f15fc92 — ThreadTrackingEmitter now exposes an onSend callback, and the initialize-from-background test fulfills its expectation from that callback instead of a hardcoded delay.

ios/Views/CardView/CardComponentViewProxy.swift
```diff
@@ -60,6 +60,12 @@ public final class CardComponentViewProxy: UIStackView {
     }
 
     @objc public func dispose() {
+        ensureMainThread { [weak self] in
+            self?.disposeOnMainThread()
+        }
```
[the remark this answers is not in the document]
descorp: Fixed in commit https://github.com/Adyen/adyen-react-native/commit/214d061f1d1741bd42553d274e7397f85f15fc92 — dispose(): now captures self strongly in the ensureMainThread block so cleanup always completes even if the proxy is deallocated right as dispose() is called (the closure is transient, not retained, so this does not introduce a retain cycle).

ios/Views/CardView/CardComponentViewProxy.swift
```diff
@@ -60,6 +60,12 @@ public final class CardComponentViewProxy: UIStackView {
     }
 
     @objc public func dispose() {
+        ensureMainThread {
```
nauaros, line 63: Should we `weakify` self here just in case

ios/Views/CardView/CardComponentViewProxy.swift
```diff
@@ -81,6 +87,12 @@ public final class CardComponentViewProxy: UIStackView {
     // MARK: - Component initialization
 
     private func initializeComponentIfNeeded() {
+        ensureMainThread { [weak self] in
+            self?.initializeComponentIfNeededOnMainThread()
+        }
+    }
+
+    private func initializeComponentIfNeededOnMainThread() {
```
erenbesel, line 95: Not long enough name haha. I think this function's name doesn't need the `onMainThread` part. It's not checking for main thread itself anyway. You call it inside `ensureMainThread` that covers it.

descorp: This part I am also unhappy with. @atmamont set this trend in previous PR and now it is codebase-wide style.
I'll address this in refactoring PR #1163 .

ios/Views/CardView/CardComponentViewProxy.swift
descorp: This is an interesting one. 

It was `[weak]` originally but AI suggested to keep it strong referenced. Idea behind it: it guarantees view will be unregistered. 

- With `[weak self]`, a background-thread `dispose()` followed by an immediate release means `self == nil` when the main-queue block runs, so the view stays registered in the bus forever.
- `dispose()` is a terminal teardown. With a strong `self`, `unregister(viewId:)` always executes, even if the caller drops its reference the instant after calling `dispose()`.
- No retain cycle. The closure is transient: passed to `ensureMainThread`, executed once, then released. It's not stored on `self`, so `self` deallocates right after the block finishes.

ios/Views/CardView/CardComponentViewProxy.swift
atmamont: Vlad what trend?

ios/Views/CardView/CardComponentViewProxy.swift
nauaros: Yeah, makes sense. Then in this case it's better to keep the strong reference.

nauaros (approved):

ios/Views/CardView/CardComponentViewProxy.swift
descorp: Naming methods "*OnMainThread" :)
It started in c143c27

erenbesel: The function does no main thread checks so there is no point to have mainThread on its name right? Rename here makes sense IMO

erenbesel: Oh, that seems like a misunderstanding to me. Does not make sense to name a function with main thread when inside the function no main thread operation is done. I'd suggest renaming all of those that does not switch to mainthread inside.

PR #1178, opened 2026-07-15
fix: RN example billing address visibility and iOS result-code

Robert-SD:
## Summary
- Fixes the React Native example app so that the billing address visibility set in `index.netlify.js` (`addressVisibility: 'none'`) is preserved when a feature config is injected via `ExternalConfigProvider`.
- Removes the `?? 'lookup'` fallback in `checkoutConfiguration.ts` so the SDK/parser defaults are used when no address visibility is configured.
- Adds `'**/build/'` to `eslint.config.mjs` ignores to avoid linting generated build artifacts.
- Adds `testID="result-code"` to `ResultView` so the iOS E2E test can identify the result screen.
- Updates the checkout completion handling to support direct session result flows (`apiClient.usesDirectSessionResult`).
- Updates iOS project files and `Podfile.lock` from the latest `pod install`.

## Test plan
- [x] `yarn typecheck`
- [x] `npx eslint` on changed files
- [x] `./gradlew :adyen_react-native:test`
- [x] React Native Android E2E tests passed (2/2)

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -34,6 +34,6 @@ export default defineConfig([
     },
   },
   {
-    ignores: ['node_modules/', 'lib/'],
+    ignores: ['node_modules/', 'lib/', '**/build/'],
   },
 ]);
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -34,6 +34,6 @@ export default defineConfig([
     },
   },
   {
-    ignores: ['node_modules/', 'lib/', '**/build/'],
+    ignores: ['node_modules/', 'lib/'],
   },
 ]);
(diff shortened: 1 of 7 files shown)


eslint.config.mjs
```diff
@@ -34,6 +34,6 @@ export default defineConfig([
     },
   },
   {
-    ignores: ['node_modules/', 'lib/'],
+    ignores: ['node_modules/', 'lib/', '**/build/'],
```
descorp: Why?
```suggestion
    ignores: ['node_modules/', 'lib/'],
```

descorp (approved):

